PR #380, opened 2013-11-12
Added level for as default parameter when creating the client

olofom:
Adds the possibility to specify log level when creating the client (by sending in logging.INFO/WARNING etc)


xordoquy:
Sorry, I'm running out of time.
I restart the travis build. It should get rid of whatever issue there was before

PR #442, opened 2014-03-26
add oneway options to speed up gevent async send

lxyu:
ignore response when `oneway=true` added to dsn.


dcramer:
Closing this as there's no clear reason its beneficial

PR #516, opened 2014-10-29
Handle streaming requests.

jmcarp:
In Tornado RequestHandler classes decorated with the stream_request_body
decorator, the request body may be a Future instance rather than a
string, which breaks currently breaks JSON serialization. This patch
uses a modified JSON encoder for the Tornado module to handle this case
and adds a test to prevent regressions.

### DIFF
--- a/raven/contrib/tornado/__init__.py
+++ b/raven/contrib/tornado/__init__.py
@@ -8,11 +8,27 @@ raven.contrib.tornado
 from __future__ import absolute_import
 
 import time
+import uuid
+import zlib
+import base64
+import datetime
 
 import raven
 from raven.base import Client
-from raven.utils import get_auth_header
+from raven.utils import json, get_auth_header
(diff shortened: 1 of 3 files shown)


raven/contrib/tornado/__init__.py
```diff
@@ -8,11 +8,27 @@
...
+
+
+class TornadoJSONEncoder(json.BetterJSONEncoder):
+    ENCODER_BY_TYPE = {
+        uuid.UUID: lambda o: o.hex,
+        datetime.datetime: lambda o: o.strftime('%Y-%m-%dT%H:%M:%SZ'),
+        set: list,
+        frozenset: list,
+        bytes: lambda o: o.decode('utf-8', errors='replace'),
+        Future: lambda o: None,
```
chrisseto: Perhaps you could do something slightly more explicit about Futures rather than just None.
python```
if future.resolved():
    return yield from future
else:
    return future.__repr__()
```

Other than that :shipit:

PR #531, opened 2014-11-22
Sanitize extra data passed into sentry

jbarbuto:
The need for this arose through use of the celery integration, which [passes task arguments as extra data](https://github.com/getsentry/raven-python/blob/5.1.1/raven/contrib/celery/__init__.py#L38). The [SanitizePasswordsProcessor docstring](https://github.com/getsentry/raven-python/blob/5.1.1/raven/processors.py#L64]) implies this is expected behavior.


raven/processors.py
```diff
@@ -34,6 +34,9 @@ def process(self, data, **kwargs):
         if 'request' in data:
             self.filter_http(data['request'])
 
+        if 'extra' in data:
+            self.filter_extra(data)
```
xordoquy: Actually, it ought to be data['password'] = '******' in the second example.
if you want to remove it, you can pop it too.
I'm not really happy since filter_extra and filter_http would be different. The reason is, filter_extra shouldn't interact with anything that's not extra. This is possible with the current proposal.

xordoquy:
```
>>> data = {'extra': {'password': '123456'}}
>>> def filter_extra(data):
...    data['password'] = '*********'
>>> filter_extra(data['extra'])
>>> print(data)
{'extra': {'password': '*********'}}
```

jbarbuto:
Ok, I changed it so `filter_extra` returns its sanitized data. Given how `filter_extra` uses [varmap](https://github.com/getsentry/raven-python/blob/5.2.0/raven/utils/__init__.py#L32) (which itself returns sanitized data instead of modifying in place), this is the only way I could see to avoid passing in the whole data dictionary.

Test errors were linting failures in unrelated code merged from master to fix a merge conflict.

xordoquy:
@jbarbuto I'll fix master later this evening and will rerun the CI on this PR

xordoquy:
Thanks for the PR.

PR #547, opened 2015-01-04
Fix syntax error under 2.7

josegonzalez:


xordoquy:
I'm not sure this is equivalent. @dcramer thoughs ?

josegonzalez:
What was the * supposed to be doing? I've never seen that operator on it's own within a function declaration.

xordoquy:
I'd assume it's for positional arguments before keywords ones. Related to the new unpacking PEP 448 though the document doesn't mention this.

asvetlov:
Please no. It's python 3 syntax, the module is not intended to work on python 2 (asyncio and aiohttp are python3-only libraries).
Regarding to syntax warning (I assume that's the source of proposed changes) please take a look on `yield from` expression also.

Yes, you can get rid of keyword-only arguments but cannot drop or replace `yield from`

josegonzalez:
Poop, I was hoping to avoid the syntax error during install. Guess I'll close this then.

PR #548, opened 2015-01-09
Add versioning helpers

dcramer:
The idea is to support easy release mappings:

```
client = raven.Client(release=raven.fetch_git_sha('./'))
```

Or

```
client = raven.Client(release=raven.fetch_package_version('sentry'))
```

### DIFF
--- a/raven/versioning.py
+++ b/raven/versioning.py
@@ -1,0 +1,33 @@
+from __future__ import absolute_import
+
+__all__ = ('fetch_git_sha', 'fetch_package_version')
+
+import os.path
+import pkg_resources
+
+from .exceptions import InvalidGitRepository
+
+
+def fetch_git_sha(path, head='master'):
+    """
+    >>> fetch_git_sha(os.path.dirname(__file__))
+    """
+    revision_file = os.path.join(path, '.git', 'refs', 'heads', 'master')
+    if not os.path.exists(revision_file):
+        if not os.path.exists(os.path.join(path, '.git')):
+            raise InvalidGitRepository('%s does not seem to be the root of a git repository' % (path,))
+        raise InvalidGitRepository('Unable to find ref to head "%s" in repository' % (head,))
+
+    fh = open(revision_file, 'r')
+    try:
+        return fh.read().strip()
+    finally:
+        fh.close()
+
+
+def fetch_package_version(dist_name):
+    """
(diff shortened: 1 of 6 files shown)


dcramer:
fyi this is a WIP yet

dcramer:
Ok commit should be good. Unless someone has a better solution I will add the docs and merge this and we'll begin using it on getsentry.com.

joaoponceleao:
Just a note.
I got tripped by this in django today with a `module object has no attribute ...`. The version in pip is at 5.1.1 and doesn't include this module.
If the pip version is not going to be updated any time soon, it might be an idea to add a 'only in dev or github branch' note to the section of the docs.

xordoquy:
@dcramer mind to add me to the documentations ? I'll try to see if I can get versioned documentation

raven/versioning.py
```diff
@@ -0,0 +1,33 @@
...
+import pkg_resources
+
+from .exceptions import InvalidGitRepository
+
+
+def fetch_git_sha(path, head='master'):
+    """
+    >>> fetch_git_sha(os.path.dirname(__file__))
+    """
+    revision_file = os.path.join(path, '.git', 'refs', 'heads', 'master')
```
blueyed: @dcramer 
I suggest using "HEAD" here instead (via `.git/HEAD`).

This might be (at least) a sha, or a link to a ref, e.g. `ref: refs/heads/develop`.

Even better would be to use `git describe --dirty --always`, but that requires the `git` binary, which might not be a given.

blueyed: Apart from that, the current code should use `head` instead of `'master'` here already.

dcramer: I'm not a git expert, but I recall having issues with using .git/HEAD

Definitely would love input here, as I honestly have no desire to research the many ways this can fail.

(I fixed the ref to s/master/HEAD/)

blueyed: The problem with .git/HEAD is/was probably that it needs to get parsed/followed when it starts with `ref:` etc.

I found the following to be useful for me, but it requires `git` to be installed:

    subprocess.check_output(['git', 'describe', '--always', '--dirty']).rstrip()

blueyed: Let's move the discussion about this into #556.

PR #553, opened 2015-01-16
Replace obsolete web client in Twisted transport

hynek:
Raven currently uses an obsolete API to do its HTTP(S) requests.

Using the “new” Agent API adds the following features:

1. Certificate and hostname validation (current implementation is MITMable).
2. A persistent HTTP pool.

I’m not sure how much of a deal 1. is WRT to backward-compatibility (people are known to run broken TLS setups after all).

I would have written tests but turns out, that Raven has no tests for its Twisted transport.  FWIW, there’s a py.test plugin to run Twisted tests: https://pypi.python.org/pypi/pytest-twisted/ but I’m not sure if you want to add it to your suite.

### DIFF
--- a/raven/transport/twisted.py
+++ b/raven/transport/twisted.py
@@ -7,6 +7,7 @@ raven.transport.twisted
 """
 from __future__ import absolute_import
 
+import io
 import logging
 
 from raven.transport.base import AsyncTransport
@@ -14,8 +15,10 @@ from raven.transport.http import HTTPTransport
 from raven.transport.udp import BaseUDPTransport
 
 try:
-    import twisted.web.client
     import twisted.internet.protocol
+    from twisted.internet import reactor
+    from twisted.web.client import Agent, FileBodyProducer, HTTPConnectionPool
+    from twisted.web.http_headers import Headers
     has_twisted = True
 except:
     has_twisted = False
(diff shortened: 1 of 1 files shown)


xordoquy:
Thanks for the PR.
I'll likely split the various integration part away from raven later. I'll probably consider pytest-twisted once it's removed from the "core".
I'll try to take some time next week to review this.

hynek:
Oh, will the integrations be even plugable? (maybe they already are, didn’t look too closely D:)

xordoquy:
they should remain as they are, they will just be part of another package. This will allow better build matrixes since the current one is only against Django

hynek:
I’ve tweaked it a little bit, feel free to ask any questions about it general or Twisted in particular.

BTW, I’m pretty sure the UDP transport is broken (where is `twisted.internet.reactor.listenUDP` supposed to come from?).

xordoquy:
@hynek the UDP protocole will be dropped any time soon.

xordoquy:
By the way, do you know when/since what version the new agent API is available ?

hynek:
9.0, so it should be from 2009.

BTW, I just read that  raven transports should support timeout, verify_ssl, and ca_certs.  The current transport supports none of them but since those changes start verifying ssl, I guess support for it should be added before merging?

xordoquy:
@hynek most transport don't support verify_sll and ca_cert yet. Feel free to add it now or later.

xordoquy:
Thanks for the PR !

PR #555, opened 2015-01-23
Fix Twisted's transport from swallowing errors

hynek:
While working on `ssl_verify` code for the Twisted transport I noticed that currently server errors are silently ignored.

This PR fixes that, although I wasn’t sure what exception to use if the submission fails so I just used `Exception`.

Let me know, if you have any questions or suggestions, I’ll commence work on `ssl_verify` once this is merged.


xordoquy:
Sorry for the delay and thanks a lot for the pull request.

PR #558, opened 2015-02-02
Stop loading unittest(2) during normal operation

jstasiak:
On my system with CPython 2.7.3 it saves 1.1 MB (16.1 vs 17.2) of memory and 60 imports
when importing raven.handlers.logging.

I know it's negligible change in most cases but I think it's a small price for making it slightly lighter.

### DIFF
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -2,7 +2,7 @@ import fnmatch
 import os
 
 from subprocess import call
-from raven.utils.compat import TestCase
+from raven.utils.testutils import BaseTestCase as TestCase
 
 
 ROOT = os.path.normpath(
--- a/raven/utils/compat.py
+++ b/raven/utils/compat.py
@@ -46,8 +46,3 @@ except ImportError:
     from urllib import parse as _urlparse  # NOQA
 
 urlparse = _urlparse
-
-try:
-    from unittest2 import TestCase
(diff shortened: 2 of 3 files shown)


xordoquy:
looks sensible. Waiting for the build to finish.

xordoquy:
Looks like there are some places where the `TestCase` substitution is missing.

jstasiak:
Ahh, I missed one I think, let me update.

tests/functional/tests.py
```diff
@@ -2,7 +2,7 @@
 import os
 
 from subprocess import call
-from raven.utils.compat import TestCase
+from raven.utils.testutils import BaseTestCase as TestCase
```
jstasiak: Now this refactoring resulted in renaming this class twice (TestCase -> BaseTestCase -> TestCase) which I'm not a fan of but I wanted to keep the changes minimal.

xordoquy: The renaming isn't an issue to me. However, to import this we'll need to have exam installed which may not be what I'm looking for.
Anyway I can always refactor this later.

jstasiak: Yeah that's true, I assumed since this is a test file all the test dependencies will be available. I can change that if needed.

xordoquy:
Thanks for the pull request.

PR #560, opened 2015-02-04
Fix Twisted UDP transport

anjensan:
Currently TwistedUDPTransport is broken.


asvetlov:
Should raven support UDP or is it deprecated?

xordoquy:
Sorry, the UDP support is going to be dropped. Should make it clear in the documentations.

anjensan:
Ok, in this case please delete TwistedUDPTransport in next release.
Because it is not work.  Actually it never was in working state.

PR #561, opened 2015-02-05
Change suggested usage of 404CatchMiddleware

maikhoepfel:
I ran into an issue with the Sentry middleware logging 404s that we couldn't reproduce in the frontend. It turned out that one of our middlewares was working as expected by catching certain 404s, but as we followed the Raven docs, Sentry reported a 404 for it nonetheless. 
I posted a slightly more elaborate version of what's going here: https://stackoverflow.com/a/28343606/753738

### DIFF
--- a/docs/integrations/django.rst
+++ b/docs/integrations/django.rst
@@ -122,10 +122,14 @@ addition of an optional ``request`` key in the extra data::
 In certain conditions you may wish to log 404 events to the Sentry server. To
 do this, you simply need to enable a Django middleware::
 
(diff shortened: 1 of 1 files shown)


xordoquy:
Thanks. I always considered the flat page are a hack.

maikhoepfel:
Thanks Xavier.

Well, the ``LocaleMiddleware`` also sometimes catches 404s and replaces them with a redirect (that's where our issue came from). So I think it's a good change.

PR #563, opened 2015-02-07
Proxy documentation for Flask

abkfenris:
I [found](https://github.com/abkfenris/baxter-flask/issues/48#issuecomment-73361250) that when Flask was deployed behind a nginx proxy Sentry was receiving the wrong `REMOTE_ADDR` from exceptions. 

By using `ProxyFix` on the `app.wsgi_app` (with nginx forwarding the correct headers) Sentry is able to get the correct `REMOTE_ADDR` from exceptions, and I've added some documentation on how to set that up.


xordoquy:
Thanks for this.

PR #572, opened 2015-03-05
Making dsn exceptions more verbose

a-pertsev:
Exception message enriched a little to provide more information to devops who are not working with codebase, but working with configs.

### DIFF
--- a/raven/conf/__init__.py
+++ b/raven/conf/__init__.py
@@ -47,7 +47,7 @@ def load(dsn, scope=None, transport_registry=None):
     url = urlparse(dsn)
 
     if not transport_registry.supported_scheme(url.scheme):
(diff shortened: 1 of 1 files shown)


xordoquy:
I don't think it's a good idea to log the DSN as it's a private setting.
Can you elaborate on why having the DSN makes it easier to track this down ? Devops should be in charge of setting the sentry DSN so they should be aware of this already.

a-pertsev:
I agree with you, probably it was not a good idea.
We had problem with big application, which have a lot of settings and current message does not give clear point what should be fixed. We can try catch exceptions while raven is initializing and provide own message, but we need to add some arguments to current exception then. Or may be ValueError can be replaced with some ```class SentryInitException(ValueError) ``` ?

dcramer:
This will be fixed as part of #587 and the dsn-refactor

PR #575, opened 2015-03-10
sanitize sentry_dsn

graingert:


dcramer:
I wonder if we should just sanitize 'dsn' in general

xordoquy:
@dcramer likely though I'll merge this meanwhile.

xordoquy:
Thanks for the pull request @graingert

Issue #579, opened 2015-03-17
Zope contrib AttributeError getProperty on Anonymous users


puittenbroek:
Currently the code here checks is user is None or not:

Which is good, but the 'getProperty' is not always available.
Plone/Zope has a 'Anonymous User",

    (Pdb) user
    <SpecialUser 'Anonymous User'>
    (Pdb) user.__class__
    <class 'AccessControl.users.SpecialUser'>
    (Pdb) user.__class__.__module__
    'AccessControl.users'
    (Pdb) user.getProperty
    *** AttributeError: getProperty

Looking at the code, best solution would be to check if the user obj is not an instance of AccessControl.users.SpecialUser.

puittenbroek:
Or see if it has the getProperty via a hasattr. :)

xordoquy:
It seems the link to the code isn't visible.

puittenbroek:
https://github.com/getsentry/raven-python/blob/master/raven/contrib/zope/__init__.py#L98

xordoquy:
sounds fair.

puittenbroek:
Even better solution:

    from AccessControl.users import nobody # at top ofc
    if user is not None and user != nobody:

puittenbroek:
I'll make a PR tomorrow :)

xordoquy:
Perfect, looking forward to it :)

dcramer:
Closing this as the PR was merged.

PR #580, opened 2015-03-18
Zope anonymous user check

puittenbroek:
Check if the user is not actually the special Anonymous User from Zope. If it isn't it's a valid user and by that logic is authenticated, don't need to check the roles for that :)


puittenbroek:
Fixes #579

xordoquy:
@puittenbroek is the AccessControl always provided by Zope ?

puittenbroek:
@xordoquy Yes, it's a core dependency :)

xordoquy:
Thanks a lot !

puittenbroek:
You're welcome, we use raven a lot and this issue kept popping up every now and then (we use Plone, which builts on Zope). Figured I might as well fix it :D

PR #583, opened 2015-03-28
Restart threaded http transport worker after a fork

Pankrat:
Forking a process doesn't clone the threads. If the worker was already active in the parent process, the child process would be stuck with a worker with a stopped thread which wouldn't send any messages.

This is relevant when using Gunicorn in preload mode when messages are already logged in the preload phase:

https://groups.google.com/forum/#!topic/getsentry/2kRU_4lye-E

### DIFF
--- a/raven/transport/threaded.py
+++ b/raven/transport/threaded.py
@@ -35,6 +35,9 @@ class AsyncWorker(object):
         }
         self.start()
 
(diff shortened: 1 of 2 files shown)


dcramer:
Awesome! Thanks for the fix!

PR #584, opened 2015-03-28
Fix test in Python 2.6

Pankrat:
The test I've introduced doesn't work on Python 2.6 but it's easy to fix:

http://stackoverflow.com/a/19668429/63392


xordoquy:
Thanks for the fix !!

PR #588, opened 2015-03-29
Use Travis's docker builders

alex:


alex:
Bleh, `apt-get` is needed for `libevent-dev`, this can't be used yet.

xordoquy:
Unfortunately I don't think we can get faster build by tuning travis.
There are some plans to split things within raven and make the all more modular.

PR #589, opened 2015-03-30
Fixes #412 and also enable python 3.4 testing with flask

### LINKED ISSUE #412: Switch `wrap_wsgi` from defaulting to True in Flask debug mode
When I enable `debug` mode in Flask and add a valid Raven DSN, errors are reported twice to Sentry.

This is because in `debug` mode, Flask [propagates the exception](https://github.com/mitsuhiko/flask/blob/master/flask/app.py#L1395) as well as calling the [normal signal that the Sentry Flask contribution binds to](https://github.com/getsentry/raven-python/blob/master/raven/contrib/flask.py#L192). When this exception is propagated, the [Sentry middleware handles it as well](https://github.com/getsentry/raven-python/blob/master/raven/middleware.py#L37), causing the duplicated exception report.

The Sentry middleware is active because [`wrap_wsgi` defaults to True in the Flask contrib](https://github.com/getsentry/raven-python/blob/master/raven/contrib/flask.py#L93).

This functionality is confusing for new users and I believe that a solution to this problem would be:

* If Flask is not in `debug`, default `wrap_wsgi` to True
* If Flask is in `debug`, defaults `wrap_wsgi` to False
* As would be expected, always follow explicit setting of `wrap_wsgi`

If this sounds good, I'd be happy to open up a PR with the change.

svenstaro:


xordoquy:
Actually the tox settings has been out of date since the travis build was totally refactored.
That put appart, I'm glad we can get flask tested against Python 3.x.
I'm waiting for the build to finish and I'll merge it.

svenstaro:
Flask apparently dropped support for Python 3.2. I will fix that.

xordoquy:
I don't think it ever had.

svenstaro:
There you go.

xordoquy:
Thanks !!

PR #593, opened 2015-04-06
passing loglevel along to handler init

gterzian:
Hello there, I was experiencing a situation where basically every Celery log was being sent to Sentry, so I  was hoping passing along the logging level to the Handler when it is initialised might help(as the default is NOTSET see https://github.com/getsentry/raven-python/blob/master/raven/handlers/logging.py#L51), although I'm not sure what caused the issue in the first place. Any ideas?

### DIFF
--- a/raven/contrib/celery/__init__.py
+++ b/raven/contrib/celery/__init__.py
@@ -41,7 +41,7 @@ def register_logger_signal(client, logger=None, loglevel=logging.ERROR):
 
     if logger is None:
         logger = logging.getLogger()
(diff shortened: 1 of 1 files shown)


dcramer:
Did this change any behavior for you? It definitely shouldn't as setLevel is a documented/public API.

gterzian:
@dcramer thanks for the feedback, and sorry I was confused the problem was in the loglevel we set celery to when booting in up (in the command line argument), so it didn't have anything with the code in raven. Thanks again for the excellent library :+1:

PR #603, opened 2015-04-28
json encoding for Decimals and lazy translations

samuelcolvin:
I just ran into the problem in a fairly serious situation where raven should have given me details on a error processing payments. Instead if failed because it tried to serialise a django translation string.

I don't see any docs for how to set up a testing environment or run tests so I wasn't able to run this.

### DIFF
--- a/raven/utils/json.py
+++ b/raven/utils/json.py
@@ -11,7 +11,7 @@ from __future__ import absolute_import
 import codecs
 import datetime
 import uuid
-
+from decimal import Decimal
 try:
     import simplejson as json
 except ImportError:
@@ -30,9 +30,13 @@ class BetterJSONEncoder(json.JSONEncoder):
         set: list,
         frozenset: list,
         bytes: lambda o: o.decode('utf-8', errors='replace'),
+        Decimal: repr
     }
 
     def default(self, obj):
+        if hasattr(obj, '_proxy____text_cast'):
+            # django's gettext_lazy proxy strings
+            return obj.encode('utf8')
         try:
             encoder = self.ENCODER_BY_TYPE[type(obj)]
         except KeyError:
--- a/tests/utils/json/tests.py
+++ b/tests/utils/json/tests.py
@@ -2,6 +2,7 @@
 
 import datetime
 import uuid
+from decimal import Decimal
 from raven.utils.testutils import TestCase
 
 from raven.utils import json
@@ -32,3 +33,7 @@ class JSONTest(TestCase):
 
         obj = Unknown()
         assert json.dumps(obj) == '"Unknown object"'
+
+    def test_decimal(self):
+        d = {'decimal': Decimal('123.45')}
+        assert json.dumps(d) == '{"decimal": "Decimal(\'123.45\')"}'
--- a/tests/contrib/django/tests.py
+++ b/tests/contrib/django/tests.py
@@ -21,7 +21,8 @@ from django.core.signals import got_request_exception
 from django.core.handlers.wsgi import WSGIRequest
 from django.http import QueryDict
 from django.template import TemplateSyntaxError
-from django.test import TestCase
+from django.test import TestCase, SimpleTestCase
+from django.utils.translation import gettext_lazy
 
 from raven.base import Client
 from raven.contrib.django.client import DjangoClient
@@ -32,7 +33,7 @@ from raven.contrib.django.middleware.wsgi import Sentry
 from raven.contrib.django.templatetags.raven import sentry_public_dsn
 from raven.contrib.django.views import is_valid_origin
 from raven.utils.serializer import transform
-from raven.utils import six
+from raven.utils import six, json
 from raven.utils.six import StringIO
 
 from django.test.client import Client as TestClient, ClientHandler as TestClientHandler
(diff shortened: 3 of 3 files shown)


tests/utils/json/tests.py
```diff
@@ -31,4 +33,12 @@ def __repr__(self):
...
-        assert json.dumps(obj) == '"Unknown object"'
\ No newline at end of file
+        assert json.dumps(obj) == '"Unknown object"'
+
+    def test_decimal(self):
+        d = {'decimal': Decimal('123.45')}
+        assert json.dumps(d) == '{"decimal": "123.45"}'
+
+    def test_translation(self):
+        d = {'lazy_translation': gettext_lazy('testing')}
```
samuelcolvin: I think this might still fail with `ImproperlyConfigured` from django as no settings are configured.

Maybe needs moving to django tests?

tests/utils/json/tests.py
```diff
@@ -31,4 +33,12 @@ def __repr__(self):
                 return 'Unknown object'
 
         obj = Unknown()
-        assert json.dumps(obj) == '"Unknown object"'
```
dcramer: this looks like a mistake

dcramer:
Things should never fail on serialization. We do a big try/catch on each object.

Do you have the error?

tests/utils/json/tests.py
samuelcolvin: no mistake just github displaying fact my file has a newline character at the end of it in a weird way. View the file and this line is there.

raven/utils/json.py
```diff
@@ -11,7 +11,8 @@
 import codecs
 import datetime
 import uuid
-
+from decimal import Decimal
+from django.utils.functional import Promise as DjangoPromise
```
dcramer: this will actually def be a problem since raven isn't bound to Django

samuelcolvin:
Couldn't agree more, but I wasn't sure where to put the catch. this is with `raven==5.0.0`, I haven't checked if there's a catch in the newer version.

The original error:

      File "/app/.heroku/python/lib/python2.7/site-packages/raven/base.py", line 507, in capture
    Unable to process log entry: <django.utils.functional.__proxy__ object at 0x7f137a9abfd0> is not JSON serializable
        self.send(**data)
      File "/app/.heroku/python/lib/python2.7/site-packages/raven/contrib/django/models.py", line 172, in         
      File "/app/.heroku/python/lib/python2.7/site-packages/raven/base.py", line 655, in captureException
      File "/app/.heroku/python/lib/python2.7/site-packages/raven/contrib/django/client.py", line 160, in capture
        result = super(DjangoClient, self).capture(event_type, **kwargs)
      File "/app/.heroku/python/lib/python2.7/site-packages/raven/contrib/django/client.py", line 180, in send
        return super(DjangoClient, self).send(**kwargs)
      File "/app/.heroku/python/lib/python2.7/site-packages/raven/base.py", line 587, in send
        message = self.encode(data)
      File "/app/.heroku/python/lib/python2.7/site-packages/raven/base.py", line 622, in encode
        return base64.b64encode(zlib.compress(json.dumps(data).encode('utf8')))
      File "/app/.heroku/python/lib/python2.7/site-packages/raven/utils/json.py", line 52, in dumps
        return json.dumps(value, cls=BetterJSONEncoder, **kwargs)
      File "/app/.heroku/python/lib/python2.7/json/__init__.py", line 250, in dumps
        sort_keys=sort_keys, **kw).encode(obj)
      File "/app/.heroku/python/lib/python2.7/json/encoder.py", line 207, in encode
        chunks = self.iterencode(o, _one_shot=True)
      File "/app/.heroku/python/lib/python2.7/json/encoder.py", line 270, in iterencode
        return _iterencode(o, 0)
      File "/app/.heroku/python/lib/python2.7/site-packages/raven/utils/json.py", line 39, in default
        return super(BetterJSONEncoder, self).default(obj)
      File "/app/.heroku/python/lib/python2.7/json/encoder.py", line 184, in default
        raise TypeError(repr(o) + " is not JSON serializable")
    TypeError: <django.utils.functional.__proxy__ object at 0x7f137a9abfd0> is not JSON serializable

raven/utils/json.py
samuelcolvin: Shall I move back the the original solution in my first commit?

dcramer:
I think ideally what we do is two things:

1. It's fine to support decimal but you lose information by converting it to a str (we generally want reprs of values)
2. We should move the lazy test into the django suite somewhere, and ideally we can solve it in the serialization path instead of the json path

samuelcolvin:
fixed as requested.

dcramer:
I'm going to need a repro, as I removed the code that shouldnt exist and nothings failing.

dcramer:
Just for some clarity, JSON isn't intended to handle encoding of rich objects here. We do some basics to make life a bit easier, but all rich encoding is done via serializers. We have a PromiseSerializer in Django that actually fully works unmodified (I confirmed with a new test).

The Decimal behavior is also already pre-existing.

samuelcolvin:
so if I update to the latest version of raven the problem above would be fixed?

dcramer:
It's possible, but I can't easily confirm

Issue #604, opened 2015-04-29
Request Entity Too Large


pehrlich:
Recently started getting the follow error, without any explicit change to Raven or Sentry configuration.  Google results are unrevealing.  Is there something known that can cause this?  

Any proper ways of logging more fully the original `AttributeError`, which caused the Sentry error, besides disabling Sentry?

``` bash
INFO:werkzeug:172.17.42.1 - - [29/Apr/2015 01:02:51] "POST / HTTP/1.1" 500 -
ERROR:sentry.errors:Unable to reach Sentry log server: HTTP Error 413: Request Entity Too Large (url: https://app.getsentry.com/api/39671/store/, body: b'<html>\r\n<head><title>413 Request Entity Too Large</title></head>\r\n<body bgcolor="white">\r\n<center><h1>413 Request Entity Too Large</h1></center>\r\n<hr><center>nginx</center>\r\n</body>\r\n</html>\r\n')
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/site-packages/raven/transport/threaded.py", line 159, in send_sync
    super(ThreadedHTTPTransport, self).send(data, headers)
  File "/usr/local/lib/python3.4/site-packages/raven/transport/http.py", line 49, in send
    ca_certs=self.ca_certs,
  File "/usr/local/lib/python3.4/site-packages/raven/utils/http.py", line 62, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/local/lib/python3.4/urllib/request.py", line 461, in open
    response = meth(req, response)
  File "/usr/local/lib/python3.4/urllib/request.py", line 571, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/local/lib/python3.4/urllib/request.py", line 499, in error
    return self._call_chain(*args)
  File "/usr/local/lib/python3.4/urllib/request.py", line 433, in _call_chain
    result = func(*args)
  File "/usr/local/lib/python3.4/urllib/request.py", line 579, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 413: Request Entity Too Large
ERROR:sentry.errors:Failed to submit message: "AttributeError: 'dict' object has no attribute 'encode'"
```

dcramer:
Is this getsentry.com or are you hosting it yourself?

pehrlich:
Getsentry.com

dcramer:
[...]

PR #605, opened 2015-04-30
Improve failure handling and logging

dcramer:
This improves logging in a couple of places:

- Add the full data output on a hard (non-API) failure
- Attempt to reconstruct a simplified traceback when an event cannot be logged

It also moves unlogged events to the 'sentry.errors.uncaught' logger.


dcramer:
This will help with things like #604

Issue #587, opened 2015-03-29
Deprecate DSN-based transport configuration


dcramer:
I want to move all implementations to a model where the transport is explicit instead of using the less understood DSN magic. This is to standardize things across languages where sometimes the schema based model doesn't make a lot of sense.

i.e.

```
Raven(dsn='...', transport=ThreadedTransport)
```

This also means we dont have to load all these modules in-memory and it's less confusing to end-users on how to use your own.

dcramer:
I'm going to take care of this, starting with a refactor of how we manage DSNs.

xordoquy:
Sounds sensible.

kainz:
What about users of the SENTRY_DSN environment variable for environment-based configuration management?

dcramer:
@kainz you cant configure most of the clients with just the environment, and that's intended for sensitive things less so than "heres how my app works"

dcramer:
Work is being done in dsn-refactor branch.

I have a few goals with this branch and intend to rebase+force-push, so everyone should avoid touching it.

PR #609, opened 2015-05-13
Refactor DSN and transport configuration

dcramer:
Passing a transport via a scheme prefix is now deprecated in favor of explicit class loading:

```python
>>> Client(dsn, transport=HttpTransport)
```

Additionally DSN-related configuration is now held in the RemoteConfig helper, which is bound on Client.remote.

This removes raven.conf.load but otherwise should be backwards compatible.

Fixes GH-587

[Review on Reviewable](https://reviewable.io/reviews/getsentry/raven-python/609)


dcramer:
There will be a followup backwards incompatible change that removes binding of Transports to a URL. They will become generic reusable for any HTTP connection. It will only cause problems for people with custom transports (which is likely very few people).

PR #610, opened 2015-06-03
Sanitize access_token values

tgwizard:
Some services that are OAuth2 resource servers accept access tokens in the form of the `access_token` query parameter. They should not be sent to sentry.

I also added `venv` to `.gitignore`, as I usually do

```
virtualenv venv
source venv/bin/activate
```

to have the virtualenv close to the repository I'm working on. This seems a common workflow. I can rebase to remove this commit if you want.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ include/
 lib/
 .idea
 .eggs
+venv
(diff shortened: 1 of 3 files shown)


dcramer:
Fixing this upstream in sentry@ffab219

tgwizard:
Cool, thanks @dcramer. Just out of curiosity, what is the difference between https://github.com/getsentry/raven-python and https://github.com/sentry/raven-python?

dcramer:
@tgwizard was a mis-link on my part

tgwizard:
@dcramer OK, thanks :smile:

PR #617, opened 2015-06-29
Add support to 'application/json' Content-Type when reporting to Sentry

RobGThai:
I had this patch in my child class to report JSON to Sentry. Thought it would be useful to have it support inside the Sentry class itself.

### DIFF
--- a/raven/contrib/flask.py
+++ b/raven/contrib/flask.py
@@ -171,13 +171,12 @@ class Sentry(object):
 
     def get_http_info(self, request):
         """
-        Redirect HTTP info to get data from request.data instead.
+        Determine how to retrieve actual data by using request.mimetype.
         """
-        headers = dict(get_headers(request.environ))
 
         retriever = self.get_form_data
 
-        if self.is_json_type(headers['Content-Type']):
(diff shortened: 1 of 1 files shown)


dcramer:
Is request.data a common thing or is that specific to something like django-rest-framework?

RobGThai:
I'm using Flask, so I don't think it's specific to Django. Might be REST-api-framework though.

dcramer:
Doh, I totally didn't pay attention to the file changed here

mitsuhiko:
One thing I would change is this:

```python
        headers = dict(get_headers(request.environ))
        if self.is_json_type(headers['Content-Type']):
```

That makes no sense in Flask, there is already a better way to get the mimetype (that's the content type without parameters):

```python
        if self.is_json_type(request.mimetype):
```

RobGThai:
@mitsuhiko Thanks, I'm pretty new to Python so I didn't realise there's that option :bow: 

Updated.

dcramer:
Thanks!

mitsuhiko:
The default value to the function won't work.  It will be an unbound function.  Just set it to `None` instead.

PR #620, opened 2015-07-07
lets try using the container based infrastructure

joshk:


dcramer:
this will fail due to a sudo call:

```
sudo apt-get install libevent-dev
```

joshk:
Yep, give me a day, I'll get this to work :)

dcramer:
Looks to work. Thanks!

joshk:
Oh, awesome, well, I was also going to send a PR to change to using Travis for caching but wanted to check with you first what dirs you need to cache?

dcramer:
Will push up some changes for it. Is there a recommended way to cache $VIRTUAL_ENV? Can we use env vars in the cache directories?

joshk:
I see you are 10 steps ahead of me!
What do you want to cache in $VIRTUAL_ENV?

dcramer:
@joshk possibly the whole thing

joshk:
Is there a particular reason you want to cache the entire VirtualEnv? I think this will explode the cache and not really improve anything, I mean, you are already at 1-2min test runs.

dcramer:
Thats fair. Going to ignore it for now.

PR #624, opened 2015-07-11
Fix ThreadedHTTPTransport import path

hannosch:
As the title says.


dcramer:
Thanks!

dcramer:
(Doh can't merge from phone. Will grab later)

PR #628, opened 2015-07-21
Fix grammar issue in Django docs

hzy:
remove "(if not the) Python's"


dcramer:
tbh not even sure why we care about saying its popular, but this is an improvement so thanks!

PR #629, opened 2015-07-21
Respect wrap_wsgi arg passed to flask.Sentry constructor

kelseyfrancis:
Any wrap_wsgi arg passed to raven.contrib.flask.Sentry() currently gets overwritten within init_app.


dcramer:
Thanks!

PR #630, opened 2015-08-02
Implement a ClosingIterator for wsgi middleware

mattrobenolt:
I've manually verified this behavior being correct with both python 2 and 3 running in gunicorn, and it does correctly make sure that `close()` is always called, even when an error is raised while iterating.

/cc @mitsuhiko

### DIFF
--- a/raven/middleware.py
+++ b/raven/middleware.py
@@ -9,6 +9,37 @@ from __future__ import absolute_import
 
 from raven.utils.wsgi import (
     get_current_url, get_headers, get_environ)
+from raven.utils.six import Iterator, next
+
+
+class ClosingIterator(Iterator):
+    """
+    An iterator that is implements a ``close`` method as-per
+    WSGI recommendation.
+    """
+    def __init__(self, sentry, iterable, environ):
+        self.sentry = sentry
+        self.environ = environ
+        self.iterable = iter(iterable)
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        try:
+            return next(self.iterable)
+        except StopIteration:
+            # propagate up the normal StopIteration
+            raise
+        except Exception:
+            # but capture any other exception, then re-raise
+            self.sentry.handle_exception(self.environ)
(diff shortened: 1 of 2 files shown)


tests/middleware/tests.py
```diff
@@ -60,7 +79,7 @@ def test_captures_error_in_iteration(self):
             response = list(response)
 
         # TODO: this should be a separate test
-        self.assertTrue(iterable.closed, True)
+        # self.assertTrue(iterable.closed, True)
```
mattrobenolt: Here's where I was a little unclear on the behavior. It seems that when the iterator is exhausted, it shouldn't close itself, but rely on being closed by the consumer.

Is this correct? If so, this assertion should be removed instead of commented out.

raven/middleware.py
```diff
@@ -9,6 +9,37 @@
...
+        except StopIteration:
+            # propagate up the normal StopIteration
+            raise
+        except Exception:
+            # but capture any other exception, then re-raise
+            self.sentry.handle_exception(self.environ)
+            raise
+
+    def close(self):
+        self.sentry.client.context.clear()
```
mattrobenolt: This is then working under the assumption below, that it's the consumer's job to `close` the iterator, which makes sense then, and will produce undesired results if the consumer doesn't close it. But that seems par for the course in expectations of this method existing.

tests/middleware/tests.py
mitsuhiko: That is correct.

raven/middleware.py
mitsuhiko: Forgetting to close is a common problem with shitty middleware unfortunately but there is nothing we can do. We could have a dtor on it and warn if it was not closed which helps uses identify broken middleware.

mattrobenolt: What's a dtor?

JTCunning:
Closed in favor of https://github.com/getsentry/raven-python/pull/836.

PR #633, opened 2015-08-05
Don't send X-Sentry-ID when last_event_id is None

lepture:
Headers with `X-Sentry-ID: None` is useless. It is a waste of bandwidth.


dcramer:
Thanks!

lepture:
Why it fails?

PR #635, opened 2015-08-05
Only fetch context when sending exception and message

lepture:
The `before_request` will be called every request. It costs 10+ ms to collect context
information even if there is no errors.

This makes the application slow. The patch will change this behavior. It works like the tornado
plugin. It will collect context information when there is an error.

### DIFF
--- a/raven/contrib/flask.py
+++ b/raven/contrib/flask.py
@@ -217,8 +217,7 @@ class Sentry(object):
             'env': dict(get_environ(request.environ)),
         }
 
-    def before_request(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


dcramer:
I think what I'd prefer is we support passing a function to the context methods for lazy execution rather than one-offing this with each integration

pvanliefland:
@dcramer is there a solution for lazy execution now? Our app metrics show that raven can be the most time-consuming process for some requests in our Flask app (especially POST requests).

If there is no solution yet, I'd be happy to help if you can give me some pointers on how to implement it.

PR #637, opened 2015-08-08
Switch source reading to linecache

dcramer:


mitsuhiko:
Generally it would make sense to use linecache but we need to be careful because it's a bit thread unsafe.  It's usually not a problem for regular tracebacks in Python which is why you normally do not encounter this.  I  think we might have to test this under heavy load to see if it's an issue.

Small other note: linecache returns lines with newlines at the end, so we will have to strip them.

PR #639, opened 2015-08-11
Add documentation for the IGNORABLE_404_URLS setting.

clokep:
This adds documentation for the `IGNORABLE_404_URLS` setting, which is part of the Django integration.

I couldn't actually get the Sphinx documentation to build locally (and couldn't find any documentation on how to do that), so I just checked the rendering in GitHub's rST preview.


dcramer:
Thanks!

PR #641, opened 2015-08-15
In order to prepare the compatibility with the Django 1.9

dulaccc:
Django 1.9 is aimed to be [in alpha in september](https://code.djangoproject.com/wiki/Version1.9Roadmap). So this PR should at least stay open until September 21, 2015.

The changes are described in the [Django 1.9 release notes](https://docs.djangoproject.com/en/dev/releases/1.9/#template-loaderorigin-and-stringorigin-are-removed).

Cheers


dcramer:
Thanks!

PR #646, opened 2015-08-28
Fix Tornado error handling.

### LINKED ISSUE #601: Tornado transport should handle exceptions
If an error occurs while logging to sentry (for example rate limiting), Tornado (at least in recent versions) will complain about a future error which has not been retrieved. It will log the error to the Tornado logger, and if you have the SentryHandler configured for your loggers then you'll get a nice loop with logging going crazy.

The AsyncSentryClient and the TornadoHTTPTransport should run IOLoop.add_future with a callback that runs future.result() to retrieve any error, and print the error to stdout.

recht:
The error handling was quite broken - the exception handlers were never invoked, and they used methods which have been removed.

This fixes #601

### DIFF
--- a/raven/contrib/tornado/__init__.py
+++ b/raven/contrib/tornado/__init__.py
@@ -7,6 +7,9 @@ raven.contrib.tornado
 """
 from __future__ import absolute_import
 
+from functools import partial
+
+from tornado import ioloop
 from tornado.httpclient import AsyncHTTPClient, HTTPError
 
 from raven.base import Client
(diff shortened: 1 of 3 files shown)


dcramer:
Thanks for the fix (and tests!)

emou:
Can we expect a new release containing this fix soon?

dcramer:
@emou at some point, but we dont push out a release for every change, so the truth of it is "itll come when it comes". There's some other fixes we need to get out, so it'll likely be soon, but we need to ensure the other fixes resolve various problems first.

PR #650, opened 2015-09-04
Fix the Tornado transport, arguments were in wrong order.

recht:
Also, add tests to ensure that it actually works.

The error was introduced in #646.


dcramer:
Thanks!

PR #656, opened 2015-09-14
Better guessing for head in `fetch_git_sha`

akalipetis:
Currently, `fetch_git_sha` takes `master` branch as the default `HEAD`, while this is not true for many cases - like `develop` branches, feature branches etc.

This implementation uses a smarter default and takes the `HEAD` to use from the `.git/HEAD` file.

Also, includes a - possible typo - fix in the dependencies for tests in `setup.py` and an improved test to test this new behavior - where the previous implementation would fail.


dcramer:
Thanks!

PR #663, opened 2015-09-23
Pass the required parsed_url parameter back to HTTPTransport

dbravender:
Without this I'm getting this traceback:
```
  File ".../raven/transport/twisted.py", line 33, in __init__
    super(TwistedHTTPTransport, self).__init__(*args, **kwargs)
exceptions.TypeError: __init__() takes at least 2 arguments (1 given)
```


dcramer:
Thanks!

PR #669, opened 2015-10-15
Fix issue user_info is populated but not returned when SENTRY_USER_ATTRS is populated

kgritesh:
This pull request basically adds a single return statement in `get_user_info` method when `SENTRY_USER_ATTRS` is specified. It also fixes the `test_user` testcase to actually test if the user context in error event actually contains the specified attributes in `SENTRY_USER_ATTRS`


dcramer:
Well, not sure how we fucked that up, but thanks!

PR #672, opened 2015-10-19
Detected installed version instead of declared version when possible

noirbizarre:
This pull-request try to detect version with pkg_resources first as it is able to detected installed version instead of declared one.

Main use case is version tagged with ``egg_info -b $TAG`` which is not written in ``__version__`` but detected by pkg_resources.

Our continuous integration versionning heavily rely on this mecanism: ``__version__`` contains 0.1.dev and we release ``python setup.py egg_info -b $TAG bdist_wheel`` which produce a package with version ``0.1.dev$TAG`` but which is currently detected ``0.1.dev`` by raven/sentry.

### DIFF
--- a/raven/utils/__init__.py
+++ b/raven/utils/__init__.py
@@ -57,6 +57,16 @@ _VERSION_CACHE = {}
 
 def get_version_from_app(module_name, app):
     version = None
+
+    # Try to pull version from pkg_resource first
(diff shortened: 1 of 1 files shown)


dcramer:
I dont have a strong opinion here, but htis is fine with me.

Thanks!

noirbizarre:
Thanks !

PR #673, opened 2015-10-19
DSN configuration with environment variable

feldi88:
In many deployment environments the configuration of the DSN with environment variable is helpful and should be added to the documentation.


dcramer:
Thanks!

PR #677, opened 2015-10-27
Use cached_property for getting client in the Django middleware

mattrobenolt:
Since we need to lazy import the client inside the `@property`, we incur
the overhead of importing every time.

This overhead is minimal, ~0.7usec on my machine, but this is done on
every request, 3 times per request which adds up to ~2.1usec.

As a `@cached_property`, this gets transformed directly into the __dict__
of the instance, and brings the time down to ~0.04usec per lookup, which
is over an order of magnitude faster after the initial call.

Also, worth noting, `cached_property` was added in Django 1.4, and 1.4 is the oldest we test against in Travis, so I assume this is the oldest that we support. If not, I can `try...except ImportError` and fallback.

### DIFF
--- a/raven/contrib/django/middleware/wsgi.py
+++ b/raven/contrib/django/middleware/wsgi.py
@@ -7,13 +7,14 @@ raven.contrib.django.middleware.wsgi
 """
 from __future__ import absolute_import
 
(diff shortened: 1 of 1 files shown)


dcramer:
lgtm

PR #678, opened 2015-10-30
Handle non-text dictionary keys when sanitizing data.

merll:
The sanitizing process assumes that keys being checked are string types. There can however be dictionaries with integers or other immutable types (e.g. tuples) as keys. These lead to exceptions such as
```
AttributeError("'int' object has no attribute 'lower'",) 
```

This PR converts keys before proceeding (e.g. calling ``lower()``). It also includes a test extension for a nested dictionary.


dcramer:
Thanks!

Issue #675, opened 2015-10-27
Proposal: Log SystemExit in wsgi wrapper


mattrobenolt:
When using gunicorn, at least, and a worker times out, gunicorn sends a `SIGABRT` signal to the worker process. This manifests itself as a `SystemExit` exception that gets raised in the worker.

When this happens in Django, Django completely ignores `SystemExit` and lets it re-raise so the process can exit, without ever calling the `got_request_exception` signal, so in our case, we never see the exception and nobody knows that anything happened since it happily exited without a fuss.

My proposal would be to explicitly catch `SystemExit`, log, then reraise.

imo this seems like a reasonable thing to do within a middleware. If something is trying to shut down the process while in the middle of the request, probably worth logging.

In gunicorn, this wouldn't log for a normal shutdown, since that sends a `SIGTERM` instead of `SIGABRT`.

Would need to be tested against other wsgi runners, specifically uwsgi.

Refs:
https://github.com/django/django/blob/37ea3cb03e80de80380009a7a7939bc48d75abe9/django/core/handlers/base.py#L223-L225
https://github.com/benoitc/gunicorn/blob/e1912db59a2f1df36387dd28bb3515202cc03098/gunicorn/arbiter.py#L443

dcramer:
this sounds like noise to me -- how is it different than things like the client disconnecting before the response finishes streaming?

mattrobenolt:
That shouldn't trigger a `SystemExit`.

This timeout is specifically when gunicorn kills the worker because it was taking too long to respond. Maybe a busy loop or something like that. So this goes unnoticed.

If the client disconnects, it should be an IOError or something akin to that. I'd have to investigate exactly what happens in that case a bit more. But it's unlikely that it's a `SystemExit` since that's telling the process to die.

dcramer:
The problem is that "SystemExit" is not a generic error. It'd be akin to us logging an error whenever you called sys.exit(1), which doesnt make much sense. We have the syshook handlers which capture things that get bubbled up (and raise SystemExit should), so I'm not sure what case this is specifically trying to cater towards.

mattrobenolt:
For posterity:

```python
# timeout.py
from time import sleep

def application(env, start_response):
    try:
        sleep(5)
    except:
        import traceback
        traceback.print_exc()
        raise
```

Run with `gunicorn timeout:application -t 1`

mattrobenolt:
[...]

PR #686, opened 2015-11-13
Log SystemExit in wsgi middleware when code != 0

mattrobenolt:
SystemExit does not subclass Exception, so we don't catch this.
SystemExit manifests itself (at least) from gunicorn when gunicorn
attempts to gracefully shut down a worker process. See GH-675 for more
background.

Fixes: GH-675

### DIFF
--- a/raven/middleware.py
+++ b/raven/middleware.py
@@ -36,12 +36,20 @@ class Sentry(object):
         except Exception:
             self.handle_exception(environ)
             raise
+        except SystemExit as e:
+            if e.code != 0:
+                self.handle_exception(environ)
+            raise
 
         try:
             for event in iterable:
                 yield event
         except Exception:
             self.handle_exception(environ)
+            raise
+        except SystemExit as e:
+            if e.code != 0:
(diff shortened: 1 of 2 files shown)


dcramer:
This might cause some duplicate events, but otherwise seems fine

mattrobenolt:
In what cases do you think it'd cause duplicate events?

dcramer:
If this propagates up to the system hook

mattrobenolt:
Ah, lemme test that out and see what actually happens there. Not sure if the system hook ignores a `SystemExit` or not. My hunch is "yes", considering these errors weren't getting captured before.

mattrobenolt:
```python
import sys

def hook(*exc_info):
    print 'going down'

sys.excepthook = hook

raise SystemExit(1)
```

`excepthook` ignores `SystemExit`

mitsuhiko:
Merged and also log `KeyboardInterrupt` which is currently also not handled.  This serves a similar purpose for other servers.

PR #693, opened 2015-11-21
Modify client.capture_exceptions to return both decorator and context manager

HPotter:
Client.capture_exceptions is useful, but sometimes it may be handy to have a context manager. So I decided not to add new methods, but extend capture_exceptions functionality using contextlib.contextmanager, which acts as both a decorator and a context manager

### DIFF
--- a/raven/base.py
+++ b/raven/base.py
@@ -17,9 +17,13 @@ import uuid
 import warnings
 
 from datetime import datetime
-from functools import wraps
 from pprint import pformat
 from types import FunctionType
+
+if sys.version_info >= (3, 2):
+    import contextlib
+else:
+    import contextlib2 as contextlib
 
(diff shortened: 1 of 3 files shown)


dcramer:
Thanks!

felixonmars:
@HPotter Why not allow contextlib for Python 2.5+ too? AFAIK it's also a decorator as well as a contextmanager...

felixonmars:
Nevermind, I got it... Would still be great to avoid the dependency on a library with no update since 2012, though...

dcramer:
@feldi88 are you aware of issues in contextlib? otherwise if it works without any problems, it doesnt really need updated. Not everything is an on-going iterative issue, many things are solvable and then can be left where they're at.

Issue #676, opened 2015-10-27
raven.fetch_git_sha doesn't support detached head checkouts


brentc:
When a checkout is in 'detached' head mode, `.git/HEAD` contains a git commit reference. `fetch_git_sha` [doesn't handle this case](https://github.com/getsentry/raven-python/blob/a93f8201fe08af37d204153bad3dbb20c11ca973/raven/versioning.py#L30-L35).

We'd like to use this feature for our codebase, but our deploy process currently uses this type of checkout.

```
$ git checkout --force ea0bfa7b83b5a6b816a1da0360bde4950a73fa28
Note: checking out 'ea0bfa7b83b5a6b816a1da0360bde4950a73fa28'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b new_branch_name

HEAD is now at ea0bfa7... [omitted]

python -c 'import raven; raven.fetch_git_sha(".")'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/brent/virtualenvs/Kira-Talk/lib/python2.7/site-packages/raven/versioning.py", line 42, in fetch_git_sha
    raise InvalidGitRepository('Unable to find ref to head "%s" in repository' % (head,))
raven.exceptions.InvalidGitRepository: Unable to find ref to head "ea0bfa7b83b5a6b816a1da0360bde4950a73fa28" in repository
```

tvavrys:
+1

Opa-:
:+1:

PR #694, opened 2015-11-21
Support detached HEAD in fetch_git_sha

ovidner:
Quick fix for #676. Unsure if this might break some other case – but it seems to work when the HEAD contains `ref: ...` or an SHA.

### DIFF
--- a/raven/versioning.py
+++ b/raven/versioning.py
@@ -32,7 +32,7 @@ def fetch_git_sha(path, head=None):
                 path, '.git', *head.rsplit(' ', 1)[-1].split('/')
             )
         else:
-            revision_file = os.path.join(path, '.git', head)
+            return head
     else:
         revision_file = os.path.join(path, '.git', 'refs', 'heads', head)
 


dcramer:
This definitely seems like it'd cause any number of problems if it was 'ref: master' or similar. What did you mean by seems to work?

andreif:
@dcramer This fix is correct. `ref: ...` is handled in the first part of the `if` statement. 
There is no such path `.git/:sha:` anyway

dcramer:
nvm not sure how I missed that

dcramer:
Thanks!

ismailsunni:
Hi 
I am using the latest version of raven (version 6.6.0), but I still got the error:
```
  File "/usr/local/lib/python3.6/site-packages/raven/versioning.py", line 25, in fetch_git_sha
    'Cannot identify HEAD for git repository at %s' % (path,))
raven.exceptions.InvalidGitRepository: Cannot identify HEAD for git repository at 

```

Is it not included in this version? If not, which version I should use?

dcramer:
@ismailsunni you can confirm that the correct file is on disk and referenced. You could alternatively just bind it yourself if our tools aren't picking it up automatically.

ismailsunni:
hi @dcramer 
Ah, my bad... my git directory is outside the django-project, so I can't find the .git directory. Thanks replying :)

PR #698, opened 2015-12-04
enable automated testing on django 1.9 and python 3.5

nikolas:


dcramer:
We really need to sit down and fix the tests, but this is good. Thanks!

PR #706, opened 2015-12-16
Catch the RuntimeError raised if contentypes is not available

### LINKED ISSUE #705: Django 1.9 exception if contrib apps are not loaded
In Django 1.9 it became an error to import a model of an app before it is loaded (https://docs.djangoproject.com/en/1.9/releases/1.9/#features-removed-in-1-9):

>All models need to be defined inside an installed application or declare an explicit app_label. Furthermore, it isn’t possible to import them before their application is loaded. In particular, it isn’t possible to import models inside the root package of an application.

If you are running a lean django setup without users / content types the import of AbstractBaseUser / BaseUser triggers this condition and raven explodes.

````
Unable to process log entry: Model class django.contrib.contenttypes.models.ContentType doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. 
  File "/home/kieran/.virtualenvs/grizzly-services/local/lib/python2.7/site-packages/raven/contrib/django/models.py", line 175, in sentry_exception_handler
  File "/home/kieran/.virtualenvs/grizzly-services/local/lib/python2.7/site-packages/raven/base.py", line 698, in captureException
  File "/home/kieran/.virtualenvs/grizzly-services/local/lib/python2.7/site-packages/raven/contrib/django/client.py", line 142, in capture
    data.update(self.get_data_from_request(request))
  File "/home/kieran/.virtualenvs/grizzly-services/local/lib/python2.7/site-packages/raven/contrib/django/client.py", line 56, in get_data_from_request
    from django.contrib.auth.models import AbstractBaseUser as BaseUser
  File "/home/kieran/.virtualenvs/grizzly-services/local/lib/python2.7/site-packages/django/contrib/auth/models.py", line 6, in <module>
    from django.contrib.contenttypes.models import ContentType
  File "/home/kieran/.virtualenvs/grizzly-services/local/lib/python2.7/site-packages/django/contrib/contenttypes/models.py", line 159, in <module>
    class ContentType(models.Model):
  File "/home/kieran/.virtualenvs/grizzly-services/local/lib/python2.7/site-packages/django/db/models/base.py", line 103, in __new__
    "application was loaded. " % (module, name))
RuntimeError: Model class django.contrib.contenttypes.models.ContentType doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. 
```

kbrownlees:
In Django 1.9 it became an error to import a model of an app
before it is loaded (https://docs.djangoproject.com/en/1.9/releases/1.9/#features-removed-in-1-9)

This means the import of the user model may fail if the contenttype
or user contrib applications are not in INSTALLED_APPS.

Fixes #705

### DIFF
--- a/raven/contrib/django/client.py
+++ b/raven/contrib/django/client.py
@@ -52,15 +52,19 @@ class DjangoClient(Client):
         return user_info
 
     def get_data_from_request(self, request):
+        result = {}
+
         try:
             from django.contrib.auth.models import AbstractBaseUser as BaseUser
         except ImportError:
             from django.contrib.auth.models import User as BaseUser  # NOQA
-
-        result = {}
-
-        if hasattr(request, 'user') and isinstance(request.user, BaseUser):
-            result['user'] = self.get_user_info(request.user)
+        except RuntimeError:
+            # If the contenttype / user applications are not installed trying to
+            # import the user models will fail for django >= 1.9.
+            pass
+        else:
+            if hasattr(request, 'user') and isinstance(request.user, BaseUser):
+                result['user'] = self.get_user_info(request.user)
 
         try:
             uri = request.build_absolute_uri()


kbrownlees:
Catching a runtime error here is pretty nasty, but explicitly trying to check INSTALLED_APPS is probably
just as nasty, thoughts?

dcramer:
we can remove this if we come up with some better, but for now this seems ok

adepue:
:+1: for this.  Thanks.  This is blocking our upgrade to django 1.9 at this point. 
Thanks for the fix!

dcramer:
5.9.1

adepue:
Thanks!

PR #719, opened 2016-01-12
Do not capture tornado.web.HTTPErrors if their status code is not 5xx

### LINKED ISSUE #715: Tornado logs any HTTPError even if not 5xx
In Tornado 3.2.2, raven 5.7.2 using ``raven.contrib.tornado.SentryMixin``, any ``tornado.web.HTTPError`` will be sent to sentry, even if their status code is not 5xx (for example a 403 will be reported).

ojii:
Fixes #715

### DIFF
--- a/raven/contrib/tornado/__init__.py
+++ b/raven/contrib/tornado/__init__.py
@@ -11,6 +11,7 @@ from functools import partial
 
 from tornado import ioloop
 from tornado.httpclient import AsyncHTTPClient, HTTPError
+from tornado import web
 
 from raven.base import Client
 
@@ -227,6 +228,9 @@ class SentryMixin(object):
         log_exception() is added in Tornado v3.1.
         """
         rv = super(SentryMixin, self).log_exception(typ, value, tb)
+        # Do not capture web.HTTPErrors outside the 500 range.
+        if isinstance(value, web.HTTPError) and (value.status_code < 500 or value.status_code > 599):
+            return rv
         self.captureException(exc_info=(typ, value, tb))
         return rv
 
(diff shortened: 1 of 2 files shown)


raven/contrib/tornado/__init__.py
```diff
@@ -11,6 +11,7 @@
 
 from tornado import ioloop
 from tornado.httpclient import AsyncHTTPClient, HTTPError
+from tornado import web
```
mattrobenolt: Can we just do `from tornado.web import HTTPError as TornadoHTTPError` and avoid the attribute lookup?

mattrobenolt: Or uh, `WebHTTPError` would make more sense since the other `HTTPError` is also from tornado. :(

ojii: I did it this way because of ``tornado.httpclient.HTTPError`` (see line above), which isn't the same as ``tornado.web.HTTPError`` (the naming is rather confusing...). Therefore using ``web.HTTPError`` makes it clear which one is used. Maybe ``from tornado.web import HTTPError as HTTPServerError`` and change the line above to ``HTTPError as HTTPClientError`` would be better?

mattrobenolt:
This lgtm. Tornado uses http exception objects for control flow. Similar to Django's `raise Http404`.

raven/contrib/tornado/__init__.py
mattrobenolt: Yeah, that's fine. I understand the confusion. :) I had to double check the source to make sure they were indeed two separate objects at different paths and not an alias for each other.

mattrobenolt:
I'll alias this after merge. No worries. Thanks for the contribution. :cake: :ice_cream:

PR #720, opened 2016-01-12
Strip six module and rename to raven._compat

mitsuhiko:
We're seeing some odd issues with our six module interacting
badly with the six module from other libraries for some users.
The suspicion is that our six module in some cases gets unvendored.
We thus rename it to `raven._compat` (which is also what some
other Python modules do).  We still use six normally for the
tests.


raven/base.py
```diff
@@ -577,10 +580,10 @@ def _log_failed_submission(self, data):
         if 'exception' in data and 'stacktrace' in data['exception']['values'][0]:
             # try to reconstruct a reasonable version of the exception
             for frame in data['exception']['values'][0]['stacktrace']['frames']:
-                output.append('  File "%(filename)s", line %(lineno)s, in %(function)s' % {
-                    'filename': frame['filename'],
+                output.append('  File "%(fn)s", line %(lineno)s, in %(func)s' % {
```
mattrobenolt: Why'd you change this? I know the end result is the same, just curious since it's unrelated.

mattrobenolt:
Tests are broken. :(

raven/base.py
mitsuhiko: The linter kept complaining about long lines.

PR #721, opened 2016-01-12
Pin contextlib2 to 0.4.0 when env is py26

jameswnl:
Latest contextlib2 0.5.0 drops py26 support.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,8 +24,13 @@ from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
 import sys
 
+
+if sys.version_info[:2] <= (2, 6):
+    contextlib2 = "contextlib2<=0.4.0"
+else:
+    contextlib2 = "contextlib2"
 install_requires = [
-    'contextlib2',
+    contextlib2,
 ]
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -107,6 +112,7 @@ def run_tests(self):
     extras_require={
         'flask': flask_requires,
         'tests': tests_require,
+        ":python_version=='2.6'": [contextlib2_that_works_on_py26],
```
mattrobenolt: I've never seen this syntax before. Is `:python_version` some magic syntax?

mitsuhiko:
I'm in contact with Nick Coghlan to restore 2.6 support.

mitsuhiko:
Closing this, upstream works with 2.6 again: https://twitter.com/ncoghlan_dev/status/687256143446413312

setup.py
Julian: It's the (semi-undocumented) syntax for environment markers, which is what you need to use if you want wheels to properly do dependencies (otherwise you'd get the `install_requires` of the Python version that ran `pip wheel`).

Julian:
@mitsuhiko works for us. Thanks.

PR #723, opened 2016-01-14
Catch uncaught exception from thread

PierreF:
Raven use excepthook to catch uncaught exception, but this only work for main thread. If you spawn any other thread (using threading.Thread), any uncaught exception will be catched but threading.Thread and just print on stderr.

There is a known bug on Python that didn't evolved for years : http://bugs.python.org/issue1230540

On this bug, some workaround exists. This PR propose to integrate a workaround in raven client. Like the excepthook, it install an hook for thread when raven client is initialized.

It has the limitation that raven client must be initialized before creating other thread, or the hook won't be activated for that thread.

The following script show the issue: without the PR, the exception is only printed on stderr. With the PR, the exception is sent to Sentry and printed to stderr.

```python
import threading

from raven import Client

def target():
    0 / 0

Client('http://user:pass@localhost:9000/1')
t = threading.Thread(target=target)
t.start()
```

### DIFF
--- a/raven/base.py
+++ b/raven/base.py
@@ -12,6 +12,7 @@ import zlib
 import logging
 import os
 import sys
+import threading
 import time
 import uuid
 import warnings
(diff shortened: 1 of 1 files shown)


mitsuhiko:
I think this would need some fine tuning. In the current form that seems dangerous.

PR #726, opened 2016-01-19
Allow logging exclusions during object creation and `init_app`

radzinzki:
Hi there! 

This PR would allow the passage of a logging exclusions to the logging setup during the creation of the `Sentry` object or while calling `init_app`. Something like:

```python
sentry_object = Sentry(..., logging=True, logging_exclusions=("logger1", "logger2", ...))
```

This in turn would call the `setup_logging` like so:

```python
setup_logging(SentryHandler(self.client, level=self.level), exclude=the_passed_tuple)
```

Would you folks be open to something like this? If so, then i could go ahead and write some tests or clean this up a little bit.

### DIFF
--- a/docs/integrations/flask.rst
+++ b/docs/integrations/flask.rst
@@ -30,7 +30,8 @@ You can optionally configure logging too::
 
     import logging
     from raven.contrib.flask import Sentry
-    sentry = Sentry(app, logging=True, level=logging.ERROR)
+    sentry = Sentry(app, logging=True, level=logging.ERROR, \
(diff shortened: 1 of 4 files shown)


mitsuhiko:
I think I'm okay with adding this if there are tests and documentation for it.

radzinzki:
Alright then, I'll get to it.

radzinzki:
I've added tests and some documentation. How does this look now?

/cc @mitsuhiko

mitsuhiko:
Awesome, thanks!

PR #730, opened 2016-01-23
Remove unnecessary/confusing variable from Celery contrib

jamesob:
This `logger` variable is, as far I can tell, ineffectual and also confusing.

I didn't remove the kwarg from the function definition because I figure we should maintain API compatibility (despite the fact that this arg doesn't seem to do anything). Maybe we should add a deprecation warning in prep for removal?

Thanks for the great product.

### DIFF
--- a/raven/contrib/celery/__init__.py
+++ b/raven/contrib/celery/__init__.py
@@ -7,6 +7,8 @@ raven.contrib.celery
 """
 from __future__ import absolute_import
 
+import logging
+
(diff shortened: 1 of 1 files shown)


dcramer:
We could always remove it. No strong opinion either way. LMK and I'll merge.

jamesob:
Sorry for the late reply. Just figured it'd be better to remove that variable 'cause it shadows the `logger` that's actually used.

dcramer:
Thanks!

PR #734, opened 2016-02-11
Fix documentation for the gevent transport

rguillebert:
The gevent transport is called GeventHTTPTransport in the docs but its real name is GeventedHTTPTransport


mattrobenolt:
Thanks! :icecream:

PR #737, opened 2016-02-24
Dont implicitly configure log handlers

dcramer:
Because of the way logging propagation works this is complicated to ensure it's intelligent. Let's leave it up to the user to ensure LOGGING is configured appropriately for their application.


dcramer:
h/t @Kronuz for dissecting this issue

PR #746, opened 2016-03-29
Check for falsy dsn and not None dsn

mitsuhiko:
This only sets the DSN if it's not falsy. This means you can now pass the empty string to pick up the default instead of having to set it to `None`.

### DIFF
--- a/raven/base.py
+++ b/raven/base.py
@@ -187,13 +187,13 @@ class Client(object):
             self.install_sys_hook()
 
     def set_dsn(self, dsn=None, transport=None):
-        if dsn is None and os.environ.get('SENTRY_DSN'):
+        if not dsn and os.environ.get('SENTRY_DSN'):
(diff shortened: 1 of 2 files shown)


mitsuhiko:
@getsentry/api can someone review?

mattrobenolt:
We should prob add some tests for this.

mitsuhiko:
@mattrobenolt updated.

raven/context.py
```diff
@@ -7,12 +7,38 @@
...
+import time
+
 from collections import Mapping, Iterable
+from datetime import datetime
 from threading import local
 
 from raven._compat import iteritems
 
 
+class BreadcrumbBuffer(object):
```
mattrobenolt: gj

mitsuhiko: I was just testing you!

mitsuhiko:
@mattrobenolt ready to merge?

mattrobenolt:
:+1:

PR #748, opened 2016-04-12
Be more explicit that Django logging integration only captures ERROR

dmclain:
The django LOGGING integration docs don't mention that they only capture ERROR and a user was having trouble in IRC (https://botbot.me/freenode/sentry/2016-04-12/?msg=63974495&page=1) as a result.

It's a fine line between documenting all the minutiae of how this works and sending people to the Python logging docs, but I imagine the most common modification people would want is to send WARNING or INFO to Sentry.

### DIFF
--- a/docs/integrations/django.rst
+++ b/docs/integrations/django.rst
@@ -78,8 +78,8 @@ override:
 Integration with :mod:`logging`
 -------------------------------
 
-To integrate with the standard library's :mod:`logging` module the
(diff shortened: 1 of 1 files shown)


dcramer:
I'm ok with this, but very soon we're going to say "logging events are no longer valid in Sentry" as we double down on being used for reporting issues, and not just aggregating arbitrary logs.

PR #751, opened 2016-04-14
Make sure tornado implementations have access to future

friedcell:
send_encoded not returning seems to be a simple mistake - only use returns its value and django impl returns a value.

Returning a future in AsyncSentryClient makes sense for times when you want to make sure the error got sent to Sentry or you would like to keep it somewhere so you can wait for it later (for testing purposes).


dcramer:
This seems sane to me. I dont think changing the capture API will really cause any issues for people.

PR #752, opened 2016-04-19
Catch all exceptions raised from setuptools

mattrobenolt:
Newer versions of setuptools have changed the API to be more restrictive
and whatnot, and now spits different errors.

Instead of whitelisting specific errors, just catch them all so we never
have to worry about it again.

This also had to be done in Sentry: https://github.com/getsentry/sentry/commit/d32b14020ec37b7e895f443409dd004afa67f298
@mitsuhiko


mitsuhiko:
lgtm.

PR #753, opened 2016-04-20
Buffer log messages into breadcrumbs.

mitsuhiko:
WIP

### DIFF
--- a/raven/base.py
+++ b/raven/base.py
@@ -430,6 +430,11 @@ class Client(object):
         data.setdefault('event_id', event_id)
         data.setdefault('platform', PLATFORM_NAME)
 
+        # insert breadcrumbs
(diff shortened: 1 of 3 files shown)


mitsuhiko:
Probably not the way forward. To copy what i wrote on slack:

```
we leave logging as it is completely but make the integrations stop configuring logging
[23:49] 
instead we have a separate system which hooks logging differently and only records crumbs
[23:49] 
and we enable that instead for all integrations
[23:49] 
if people still want the old logging stuff they just need to configure it themselves
[23:49] 
and all integrations we have set up patterns for useful loggers they want to turn on for the logging capturing thing
[23:50] 
eg: requests, flask's logger, django's loggers etc.
[23:50] 
that way they feed together into the same system and we don't double record
```

PR #755, opened 2016-04-22
Add settings config site and release template tags

bozdoz:
Thought it useful to keep the javascript client and the python client using the same tags (release tag is especially useful since it's automatic).


bozdoz:
I don't like this PR as much as the more flexible one I just sent: PR #756

PR #759, opened 2016-04-25
Detect disabled threads for uwsgi

mitsuhiko:


mattrobenolt:
I assume it'd be hard to fall back to a non-threaded transport in this case so it at least _does_ work rather than not working at all?

raven/utils/compat.py
```diff
@@ -46,3 +46,17 @@
...
 urlparse = _urlparse
+
+
+def check_threads():
+    try:
+        from uwsgi import opt
+    except ImportError:
+        return
+
+    if str(opt.get('enable-threads', '0')).lower() in ('false', 'off', 'no', '0'):
```
mattrobenolt: plz comment with related uwsgi source code: https://github.com/unbit/uwsgi/blob/master/core/uwsgi.c#L4010-L4019 in case these values change in the future.

mattrobenolt:
:+1:

PR #763, opened 2016-05-05
record_breadcrumb requires type as first arg

streeter:
See [the code](https://github.com/getsentry/raven-python/blob/76b29f1d41ff2e251f6630243a3a833651268463/raven/breadcrumbs.py#L59)

### DIFF
--- a/docs/breadcrumbs.rst
+++ b/docs/breadcrumbs.rst
@@ -88,7 +88,7 @@ Example:
 
     from raven.breadcrumbs import record_breadcrumb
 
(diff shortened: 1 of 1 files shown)


mattrobenolt:
Can you change this argument to `default` instead of `info`? It'd be more appropriate example.

Thanks!

streeter:
> Can you change this argument to default instead of info? It'd be more appropriate example.

Done!

mattrobenolt:
Thanks. :)

PR #765, opened 2016-05-12
Use built-in contextlib on python 2.7

nikolas:
We don't need to use the contextlib2 backport on python 2.7

### DIFF
--- a/raven/base.py
+++ b/raven/base.py
@@ -20,10 +20,10 @@ from datetime import datetime
 from pprint import pformat
 from types import FunctionType
 
-if sys.version_info >= (3, 2):
-    import contextlib
-else:
+try:
     import contextlib2 as contextlib
+except ImportError:
+    import contextlib
 
 try:
     from thread import get_ident as get_thread_ident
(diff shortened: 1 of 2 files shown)


mitsuhiko:
I think I'm fine merging this but I would like to understand if it causes any problems so that this change is necesssary.

nikolas:
It's just a minor thing, but the existing code causes an ImportError when attempting to use raven on python 2.7 without the contextlib backport, which shouldn't be necessary in python 2.7.

It looks like this was actually introduced to the stdlib in python 2.5: https://docs.python.org/2/library/contextlib.html but I'm only testing this in python 2.7 so I've limited the scope to that version.

```
      File "/var/www/uelc/releases/blue/uelc/ve/local/lib/python2.7/site-packages/raven/__init__.py", line 57, in <module>
        from raven.base import *  # NOQA
      File "/var/www/uelc/releases/blue/uelc/ve/local/lib/python2.7/site-packages/raven/base.py", line 26, in <module>
        import contextlib2 as contextlib
    ImportError: No module named contextlib2
```

nikolas:
Well, nevermind. Maybe the backport is required after all: https://travis-ci.org/getsentry/raven-python/jobs/129760466

PR #768, opened 2016-05-15
Replace culprits with transaction names

dcramer:
- Implemented in Django via middleware
- Implemented in Flask via signals
- Implemented in Celery via signals
- Defer empty culprit to server (Sentry deals fills it)
- Adds SentryDjangoHandler to abstract setup/teardown of various listeners
- Adds SentryCeleryHandler to abstract setup/teardown of various listeners

For Django and Flask this will replace the culprit with a simplified URL path e.g. ``/foo/{bar}/``. For Celery this will use the registered task name.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/768)

### DIFF
--- a/raven/contrib/django/models.py
+++ b/raven/contrib/django/models.py
@@ -16,12 +16,16 @@ import sys
 import warnings
 
 from django.conf import settings
+from django.core.signals import got_request_exception, request_started
+from threading import Lock
 
-from raven._compat import PY2, binary_type, text_type, string_types
+from raven._compat import PY2, binary_type, text_type
 from raven.utils.conf import convert_options
 from raven.utils.imports import import_string
 
 logger = logging.getLogger('sentry.errors.client')
+
+settings_lock = Lock()
 
 
 def get_installed_apps():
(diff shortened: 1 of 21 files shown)


dcramer:
I'm not sure I'm content with this yet. There's a lot of caveats in how we do things. I'm wondering if it might be better just to focus on the URL instead.

raven/contrib/django/models.py
```diff
@@ -151,66 +155,95 @@ def sentry_exception_handler(request=None, **kwargs):
...
-    if 'djcelery' in settings.INSTALLED_APPS:
-        try:
-            # Celery < 2.5? is not supported
-            from raven.contrib.celery import (
-                register_signal, register_logger_signal)
-        except ImportError:
-            logger.exception('Failed to install Celery error handler')
-        else:
+        # try:
+        #     ga = lambda x, d=None: getattr(settings, 'SENTRY_%s' % x, d)
```
dcramer: need to figure this out, but id prefer it be part of SentryCeleryHandler or not exist by default

dcramer:
FYI due to the way the request/response cycle works in Django this is not good enough to capture transaction names for unhandled exceptions.

We're going to need to implement this as a WSGI middleware, or at least components of, and hook get_response. We can fetch ROOT_URLCONF in there to pass to the resolver (if need be), and hook the entire lifecycle of that function.

dcramer:
In all my tests this seems to be working now.

PR #769, opened 2016-05-18
Relax pytest-timeout version requirement

gi11es:
Usage syntax hasn't changed in newer versions.
The existing strict version requirement is making
packaging needlessly difficult.

The minimum version is being bumped because the first
major release of pytest-timeout requires pytest
2.8.0 or newer.


mattrobenolt:
Seem comment here: https://github.com/getsentry/raven-python/pull/770#issuecomment-220146837

PR #770, opened 2016-05-18
Relax flake8 version requirement

gi11es:
Still maintain preventive check against next major version,
which I imagine was the reason for the <2.1 in the first place.

The current restrictive range is making packaging of
raven-python needlessly difficult.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ tests_require = [
     'Django>=1.4',
     'django-celery>=2.5',
     'exam>=0.5.2',
-    'flake8>=2.0,<2.1',
+    'flake8>=2.0,<2.6',
     'logbook',
     'mock',
     'nose',


mattrobenolt:
Out of curiosity, why does this make it hard for packaging? Raven doesn't actually depend on these, and it's just an extras. So packaging shouldn't need to touch these from my understanding.

And by packaging, I'm assuming you're referring to system packaging?

gi11es:
It is for Debian packaging. I'm trying to have all the tests run cleanly in the build phase.

While I can bypass what's in setup.py, it would be better for consistency not to. My earlier comment about it being more difficult is incorrect, now that I've made more progress I see that I can override whatever's in raven's source code if needed.

The newer versions of these two modules work fine. What's the upside of pinning to old versions? Surely the new ones have bug fixes.

gi11es:
No longer needed, as the build phase of the debian package doesn't run flake8.

PR #777, opened 2016-05-24
Add fingerprint to soft timeouts in Celery

dcramer:


mattrobenolt:
Hmm, why is this much different than what @tkaemming attempted to do with https://github.com/getsentry/sentry/pull/2415/files? I don't remember why that didn't work out well. I just know it needed to be reverted.

tkaemming:
This is why I write those verbose commit messages: getsentry/sentry@1a818cedc95e53f671d46c21d0c19aad79f5ee06

> This had no effect, since `SoftTimeLimitExceeded` exceptions were being caught and logged by `sentry.utils.safe.safe_execute` without being reraised.

dcramer:
@mattrobenolt this helps everyone not just Sentry, and we can improve the situation on our end (not everything is wrapped in safe_execute)

mattrobenolt:
Sorry, I meant, it _didn't_ solve our issue.

PR #779, opened 2016-05-26
Make implicit minimum Tornado version explicit

gi11es:
TornadoTransportTests.test__sending_successfully_calls_success_callback uses gen.maybe_future, which appeared in Tornado 4.0
TornadoAsyncClientTestCase.test_sending_to_unresponsive_sentry_server_logs_error and TornadoTransportTests.test__sending_with_error_calls_error_callback use gen.sleep, which appeared in Tornado 4.1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,7 @@ tests_require = [
     'pytest-django==2.9.1',
     'pytest-timeout==0.4',
     'requests',
-    'tornado',
+    'tornado>=4.1',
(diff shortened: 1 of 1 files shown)


dcramer:
This has been open a while, and no one chimed up suggesting its wrong. Given our huge amount of expertise (kidding) I'm going to merge this, and if we see issues come up we'll revisit it.

PR #780, opened 2016-05-26
Don't use simplejson

gi11es:
It behaves differently than stdlib json.

In fact, when simplejson is present, the JSONTest.test_decimal test fails, as the output is {"decimal": 123.45} instead of {"decimal": "Decimal(\'123.45\')"}

### DIFF
--- a/raven/utils/json.py
+++ b/raven/utils/json.py
@@ -11,10 +11,7 @@ from __future__ import absolute_import
 import codecs
 import datetime
 import uuid
-try:
-    import simplejson as json
-except ImportError:
-    import json
+import json
 
 try:
     JSONDecodeError = json.JSONDecodeError


gi11es:
The CI failure looks unrelated to the change. Not sure how I can get that retried?

adamchainz:
Do `git commit --amend --no-edit` to rebuild your commit with no changes and then `git push -f` to force a rebuild

dcramer:
Thanks!

broundee:
Since `simplejson` is no longer used after this change, the [serialization error handling](https://github.com/getsentry/raven-python/blob/master/raven/utils/json.py#L50) no longer works as expected, because the standard lib `json.JSONEncoder.__init__()` does not accept the `encoding` keyword argument, producing `unexpected keyword argument` error.

PR #788, opened 2016-06-09
Add formatted attribute to message events

dcramer:


dcramer:
This attribute will be discarded until 8.6 but will upgrade gracefully since invalid interface attributes are generally dropped.

PR #792, opened 2016-06-16
Add type check for sentry initialization for flask

mdamien:
To avoid an unrelated error when doing a `Sentry(app, dsn)`

(as discussed with @mitsuhiko on IRC)

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/792)


mitsuhiko:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/raven-python/792)*

mdamien:
Hi, thanks for merging :+1: 

Small thing I just noticed, there's a typo in the exception message: It's "client should an instance of Client" instead of "'should **be** an instance of Client". We can also make it more precise ("an instance of raven.base.Client").

*- if something can go wrong, it will*

PR #793, opened 2016-06-21
Added IOError to get stack info for graceful zip import errors

MSeal:
This PR addresses an issue where zip path imports can crash sentry in a manner similar to cProfile -- see comment/blame commit below IOError addition -- with an IOError instead of just an ImportError. Specifically this occurs if the source zip code is no longer accessible to the program (easy to trigger with an uncaught exception in PySpark codebases).

I added some tests to check for the capture with a fake loader object. Given the method has no expectations for the type of loader it seemed fine to just pass an object which always raises the tested exceptions. Additionally it was taking an excessive amount of scaffolding with fake errors to capture the actual loader which exhibits the behaviour, and is difficult to guarantee that loader will have a `get_source` method, which causes it to skip testing the critical path when absent.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/793)

### DIFF
--- a/tests/utils/stacks/tests.py
+++ b/tests/utils/stacks/tests.py
@@ -2,10 +2,10 @@
 from __future__ import unicode_literals
 
 import six
+import os.path
 
 from mock import Mock
 from raven.utils.testutils import TestCase
-
 from raven.utils.stacks import get_culprit, get_stack_info, get_lines_from_file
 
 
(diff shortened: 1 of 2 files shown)


tests/utils/stacks/tests.py
```diff
@@ -95,11 +95,40 @@ def test_frame_allowance(self):
...
         assert results['frames'][9]['filename'] == '9'
 
+class FailLoader():
+    '''
+    Recreating the built-in loaders from a fake stack trace was brittle.
+    This method ensures its testing the path where the loader is defined
+    but fails with known exceptions.
+    '''
+    def get_source(self, module_name):
+        if '.py' in module_name:
```
mattrobenolt: Isn't this going to cause a severe issue if you just had a module named `foo.pythings.whatever`?

mattrobenolt: Oh, haha, ignore me. This was just for tests. :)

MSeal: :)

MSeal:
Any chance for a merge or some requested changes on this PR? It's been open for a month now without actionable comments.

dcramer:
@MSeal sorry about that! We have far too many repositories, and no way yet to know when things are dangling :/

PR #796, opened 2016-06-29
Django client template_debug's source_lines is just a snippet

faulteh:
When using the Jinja2 template engine over the built-in Django one I came across this bug where get_data_from_template was itself breaking. As a result, when debugging, I would get the template error in my browser, but raven would never send the exception to my Sentry server as execution stopped in this function.

In the template_debug field of the Exception, the source_lines is actually a snippet, not the full source of the template source_lines is also list of tuples, not just a list of lines in the form of [(linenbr, 'line-detail'), (linenbr, 'line-detail'), ...]

So I generate source_lines starting with an empty list, adding blank lines until we get to the snippet, then add the snippet as a list.

Also, debug['start'] and debug['end'] didn't exist (although they do when using the Django template engine), but  I removed these lines because start and end are never used elsewhere in the function (only in the source elif branch)

Finally, if the context_line was the last line of the template, I padded source_lines with a few extra post_context does not break.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/796)

### DIFF
--- a/raven/contrib/django/utils.py
+++ b/raven/contrib/django/utils.py
@@ -23,12 +23,13 @@ def linebreak_iter(template_source):
 
 def get_data_from_template(source, debug=None):
     if debug is not None:
-        start = debug['start']
-        end = debug['end']
-        source_lines = debug['source_lines']
         lineno = debug['line']
         filename = debug['name']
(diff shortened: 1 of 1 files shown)


dcramer:
Thanks!

PR #809, opened 2016-07-22
Default Raven instance

dcramer:
This creates a default client instance at runtime which eliminates the need to do anything other than 'import raven' if SENTRY_DSN is set.

We can also leverage this to move more towards setters rather than client instantiation going forward.

@getsentry/python

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/809)


mitsuhiko:
I think we need to revisit this. In the current form this would cause lots of issues with monkeypatching that happens as part of the import then.

PR #811, opened 2016-07-24
Issue 550 exception chains

cjerdonek:
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/811)

PR #812, opened 2016-07-24
Remove six entirely and unify _compat into utils.compat

mattrobenolt:
Was kinda silly that we only used six for tests and didn't dogfood our own compat shim. This just hard removes it.

When doing this, I learned we had both a `_compat` and `utils.compat` module, so this merges `_compat` into `utils.compat`.

@getsentry/python


mattrobenolt:
I'll update this after GH-811 gets merged too since this adds a new `six` dependency that I'll vendor.

mitsuhiko:
Reviewed 37 of 37 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/raven-python/812)*

mattrobenolt:
This rebase might be fun. :(

PR #813, opened 2016-07-24
Replace py 3.2 classifier with 3.5

mattrobenolt:
@mitsuhiko We stopped testing against 3.2, so I'm assuming that we don't support it anymore? I know lots of libraries don't, so wasn't sure if that's an explicit decision or not. I can add it back if that's incorrect, but we should add 3.5 regardless.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/813)


cjerdonek:
FYI, the docs should also be updated here: https://docs.getsentry.com/hosted/clients/python/platform-support/

That page says Python 3.3 is the newest supported platform.

PR #815, opened 2016-07-28
Hardcode version in package init to avoid issues with badly installed libraries misreporting.

mitsuhiko:
@getsentry/infrastructure

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/815)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,17 @@ for m in ('multiprocessing', 'billiard'):
 
 from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
+import re
 import sys
+import ast
+
+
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -22,7 +22,17 @@
...
 from setuptools.command.test import test as TestCommand
+import re
 import sys
+import ast
+
+
+_version_re = re.compile(r'VERSION\s+=\s+(.*)')
+
+with open('raven/__init__.py', 'rb') as f:
+    version = str(ast.literal_eval(_version_re.search(
```
mattrobenolt: I assume bytes vs text doesn't matter here for python2/3?

Issue #818, opened 2016-08-02
Raven doesn't work with Django 1.10


piotr-szpetkowski:
Due to `BaseCommand.option_list` being deprecated in Django 1.8 and finally removed in 1.10 it seems like Raven is rather broken when using new version.

I'm getting following error (easily reproducible):
> AttributeError: type object 'BaseCommand' has no attribute 'option_list'

Steps to reproduce:
1. Create new virtualenv and install Django + raven.
2. Start new project.
3. Configure Raven at settings following docs [@Sentry Docs](https://docs.getsentry.com/hosted/clients/python/integrations/django/)
4. Try to fire up the test management command: `python manage.py raven test`

dcramer:
Considering dropping this functionality entirely since our sys.except_hook handles it now.

fengsi:
`5.24.0` "seems" to support Django 1.10 but test is still not working:
```
$ python manage raven test
AttributeError: type object 'BaseCommand' has no attribute 'option_list'
```

dcramer:
@fengsi ah we missed that one. We'll need to figure out how to support compat there, but at the very least the core integration will be working now.

PR #824, opened 2016-08-05
Fix behavior of raven command on Django 1.8+

dcramer:
Fixes GH-818

This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/824)

### DIFF
--- a/raven/contrib/django/management/commands/raven.py
+++ b/raven/contrib/django/management/commands/raven.py
@@ -2,7 +2,7 @@
 raven.contrib.django.management.commands.raven
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-:copyright: (c) 2010-2013 by the Sentry Team, see AUTHORS for more details
+:copyright: (c) 2010-2016 by the Sentry Team, see AUTHORS for more details
 :license: BSD, see LICENSE for more details.
 """
 from __future__ import absolute_import, print_function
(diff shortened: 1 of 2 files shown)


fengsi:
Tested but seems not working?
```
$ python manage.py raven
Usage: manage.py raven test
```
```
$ python manage.py raven test
usage: manage.py raven [-h] [--version] [-v {0,1,2,3}] [--settings SETTINGS]
                       [--pythonpath PYTHONPATH] [--traceback] [--no-color]
                       [--data DATA] [--tags TAGS]
manage.py raven: error: unrecognized arguments: test
```

dcramer:
Doh. I guess our unit test is totally wrong.

fengsi:
Looks so...
https://github.com/getsentry/raven-python/blob/master/tests/contrib/django/management/commands/test_raven.py#L17
`command` is passed in test manually resulting in `command` is always there but `add_arguments()` knows nothing about `test`...

And for `def handle(self, *args, **options):` the arg should be in `**options`.

I did a quick hack by adding a positional `test` arg, but it doesn't look consistent with previous message (also it's required to check "test" string for `test` arg, which is weird):
```
usage: manage.py raven [-h] [--version] [-v {0,1,2,3}] [--settings SETTINGS]
                       [--pythonpath PYTHONPATH] [--traceback] [--no-color]
                       [--data DATA] [--tags TAGS]
                       test
manage.py raven: error: the following arguments are required: test
```
BTW could we replace `print()` with `self.stderr.write()`?

dcramer:
This shoudl be fixed in 5.24.3.

Re: sys.stderr.write, I'm not opposed, but I dont care enough to go make the change at this point

PR #827, opened 2016-08-08
Remove message attribute

dcramer:
As of Sentry 8.6 the message attribute is no longer required, and causes some pollution of data when present (within exceptions).

This also removes some various deprecated behavior, and simplifies how "events" work.

@getsentry/python

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/827)


dcramer:
I'm thinking as part of these we make it 6.0. Maybe stuff in any other major refactors we want. I think there's a lot we could do to still clean up config/event mangling if we wanted.

Issue #599, opened 2015-04-19
Problem with wsgi middleware when running under gunicorn with gaiohttp worker


urbaniak:
Each request raises an error like this:

```
RuntimeError: Task got bad yield: b'<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">...
```

It's caused by yield in the middleware:
```
    def __call__(self, environ, start_response):
        self.client.http_context(self.get_http_context(environ))

        try:
            iterable = self.application(environ, start_response)
        except Exception:
            self.handle_exception(environ)
            raise

        try:
            for event in iterable:
                yield event
```

asvetlov:
What is your problem and how raven/sentry affects it?

urbaniak:
Got a flask application running on gunicorn+gaiohttp worker, it's working well without Sentry middleware.

When enabling raven middleware the yield (https://github.com/getsentry/raven-python/blob/master/raven/middleware.py#L42) causes an error as in previous post:

```
Apr 19 07:48:50 hostname gunicorn[30853]: [2015-04-19 07:48:50 -0400] [30876] [ERROR] Error handling request
Apr 19 07:48:50 hostname gunicorn[30853]: Traceback (most recent call last):
Apr 19 07:48:50 hostname gunicorn[30853]: File "/home/app/.virtualenv/lib64/python3.4/site-packages/aiohttp/server.py", line 240, in start
Apr 19 07:48:50 hostname gunicorn[30853]: yield from handler
Apr 19 07:48:50 hostname gunicorn[30853]: File "/home/app/.virtualenv/lib64/python3.4/site-packages/aiohttp/wsgi.py", line 132, in handle_request
Apr 19 07:48:50 hostname gunicorn[30853]: riter = yield from riter
Apr 19 07:48:50 hostname gunicorn[30853]: File "/home/app/.virtualenv/lib64/python3.4/site-packages/raven/middleware.py", line 42, in __call__
Apr 19 07:48:50 hostname gunicorn[30853]: yield event
Apr 19 07:48:50 hostname gunicorn[30853]: RuntimeError: Task got bad yield: b'<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">\n<title>404 Not Found</title>\n<h1>Not Found</h1>\n<p>The requested URL was not found on the server.  If you entered the URL manually please check your spelling and try again.</p>\n'
```

[...]

PR #836, opened 2016-08-19
Rebase closingiterator and apply changes since it was first posted.

JTCunning:
This is a resubmission of https://github.com/getsentry/raven-python/pull/630 which I have confirmed fixes https://github.com/getsentry/raven-python/issues/599.

I'll let @mattrobenolt and @mitsuhiko rehash their conversation here and close the past PR.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/836)

### DIFF
--- a/raven/middleware.py
+++ b/raven/middleware.py
@@ -7,8 +7,57 @@ raven.middleware
 """
 from __future__ import absolute_import
 
+from six import Iterator, next
+
 from raven.utils.wsgi import (
     get_current_url, get_headers, get_environ)
+
+
+class ClosingIterator(Iterator):
+    """
+    An iterator that is implements a ``close`` method as-per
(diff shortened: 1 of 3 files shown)


nate-opti:
Huge thanks for looking into this fix!! Just verified locally that this branch fixes the issue with our setup: 
```
Python 3.5.1
aiohttp (0.21.6)
gunicorn (19.5.0)
```

raven/middleware.py
```diff
@@ -7,10 +7,59 @@
 """
 from __future__ import absolute_import
 
+from six import Iterator, next
```
mattrobenolt: six is not a runtime dependency. We have vendored copied of bits we need.

mitsuhiko:
Merged this in a modified version.

PR #841, opened 2016-08-24
Move _emit extraction functionality into a method.

### LINKED ISSUE #840: Extract _emit record splitting functionality into a utility so it can be used in Sentry.
@mattrobenolt brought up the idea that we should better align Sentry's internal logging handler with the way that raven accepts records. One way to ease that alignment is to make [this codeblock](https://github.com/getsentry/raven-python/blob/ed91eef804eb1453ca625270fe1b67e2b687b89b/raven/handlers/logging.py#L107-L124) reusable.

JTCunning:
Fixes #840.

@mitsuhiko @mattrobenolt

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/841)

### DIFF
--- a/raven/handlers/logging.py
+++ b/raven/handlers/logging.py
@@ -26,7 +26,7 @@ RESERVED = frozenset((
 ))
 
 
-def extract_extra(record):
+def extract_extra(record, reserved=RESERVED):
     data = {}
 
     extra = getattr(record, 'data', None)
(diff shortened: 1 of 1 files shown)


dcramer:
I think this is fine, but in general I have avoided using raven utilities in Sentry since they should be independent deps (neither one should block the other)

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/raven-python/841#-:-KRdU_gaxBZfqzqoW-Fn:bvm5rqy)*

PR #846, opened 2016-08-26
Fix get_user_info for Django 1.10 when is_authenticated is a property

### LINKED ISSUE #845: raven fails to get user info on Django 1.10 when is_authenticated is a property
Starting from Django 1.10 `User.is_authenticated` isn't a method anymore, but a property ([release notes](https://docs.djangoproject.com/en/1.10/releases/1.10/#user-is-auth-anon-deprecation)).

This doesn't matter (yet) when using the builtin User model since it returns a `CallableBool`. When using a custom user model that already goes the new way and returns a "real" bool, raven fails to get the user info with the following exception:

```
File "/home/andi/.virtualenvs/proj/lib/python3.5/site-packages/raven/contrib/django/client.py", line 144, in get_user_info
    not user.is_authenticated():
TypeError: 'bool' object is not callable
```

andialbrecht:
This changes fixes #845.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/846)

### DIFF
--- a/raven/contrib/django/client.py
+++ b/raven/contrib/django/client.py
@@ -140,9 +140,14 @@ class DjangoClient(Client):
         install_sql_hook()
 
     def get_user_info(self, user):
-        if hasattr(user, 'is_authenticated') and \
-           not user.is_authenticated():
-            return None
+        if hasattr(user, 'is_authenticated'):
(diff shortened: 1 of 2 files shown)


dcramer:
Thanks!

PR #857, opened 2016-09-13
By default don't import 3rd-party Python modules, only stdlib ones.  …

### LINKED ISSUE #854: Please don't import eventlet even if it's available
This code has two problems:
```python
try:
    import eventlet
    try:
        from eventlet.green import urllib2 as eventlet_urllib2
    except ImportError:
        from eventlet.green.urllib import request as eventlet_urllib2
    has_eventlet = True
except:
    has_eventlet = False
```

1. Bare `except:` is evil, please catch a specific exception, ImportError in this case.
2. I am trying to get rid of eventlet in my code base, and so I have an assertion in my code: `assert 'eventlet' not in sys.modules`.  Unconditionally importing eventlet makes it fail that assertion.  Instead, could you please get the module only if it's already imported?  `if 'eventlet' in sys.modules: stuff`.

gjcarneiro:
…Fixes #854.

### DIFF
--- a/raven/transport/__init__.py
+++ b/raven/transport/__init__.py
@@ -9,13 +9,13 @@ raven.transport
 # manually import/register transports somehow
 from __future__ import absolute_import
 
-from raven.transport.base import *  # NOQA
-from raven.transport.eventlet import *  # NOQA
-from raven.transport.exceptions import *  # NOQA
-from raven.transport.gevent import *  # NOQA
-from raven.transport.http import *  # NOQA
-from raven.transport.requests import *  # NOQA
-from raven.transport.registry import *  # NOQA
-from raven.transport.twisted import *  # NOQA
-from raven.transport.threaded import *  # NOQA
-from raven.transport.tornado import *  # NOQA
+# from raven.transport.base import *  # NOQA
+# from raven.transport.eventlet import *  # NOQA
+# from raven.transport.exceptions import *  # NOQA
+# from raven.transport.gevent import *  # NOQA
+# from raven.transport.http import *  # NOQA
+# from raven.transport.requests import *  # NOQA
+# from raven.transport.registry import *  # NOQA
+# from raven.transport.twisted import *  # NOQA
+# from raven.transport.threaded import *  # NOQA
+# from raven.transport.tornado import *  # NOQA
--- a/raven/transport/registry.py
+++ b/raven/transport/registry.py
@@ -8,15 +8,15 @@ raven.transport.registry
 from __future__ import absolute_import
 
 # TODO(dcramer): we really should need to import all of these by default
-from raven.transport.eventlet import EventletHTTPTransport
+# from raven.transport.eventlet import EventletHTTPTransport
 from raven.transport.exceptions import DuplicateScheme
 from raven.transport.http import HTTPTransport
-from raven.transport.gevent import GeventedHTTPTransport
-from raven.transport.requests import RequestsHTTPTransport
+# from raven.transport.gevent import GeventedHTTPTransport
+# from raven.transport.requests import RequestsHTTPTransport
 from raven.transport.threaded import ThreadedHTTPTransport
 from raven.transport.threaded_requests import ThreadedRequestsHTTPTransport
-from raven.transport.twisted import TwistedHTTPTransport
-from raven.transport.tornado import TornadoHTTPTransport
+# from raven.transport.twisted import TwistedHTTPTransport
+# from raven.transport.tornado import TornadoHTTPTransport
 from raven.utils import urlparse
 
 
@@ -74,10 +74,10 @@ class TransportRegistry(object):
 default_transports = [
     HTTPTransport,
     ThreadedHTTPTransport,
-    GeventedHTTPTransport,
-    TwistedHTTPTransport,
-    RequestsHTTPTransport,
-    ThreadedRequestsHTTPTransport,
-    TornadoHTTPTransport,
-    EventletHTTPTransport,
+    # GeventedHTTPTransport,
+    # TwistedHTTPTransport,
+    # RequestsHTTPTransport,
+    # ThreadedRequestsHTTPTransport,
+    # TornadoHTTPTransport,
+    # EventletHTTPTransport,
 ]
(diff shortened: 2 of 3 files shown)


dcramer:
This will break a backwards compat change (where we let you configure the transport with the DSN), but as long as we mentioned that in the CHANGES I'm totally ok with it.

We also should simply remove the comments and just kill the lines (we're using version control after all).

gjcarneiro:
So I fixed the tests and cleaned up the code.  One issue that I had was that I have to access `Client._registry`, and we shouldn't be required to access attributes marked private.  So I propose to rename, `Client._registry` to `Client.registry`, what do you think?

dcramer:
Thanks!

CHANGES
```diff
@@ -3,6 +3,37 @@ Version 5.27.0
...
+  ```
+  from raven.transport.eventlet import EventletHTTPTransport
+  from raven.transport.gevent import GeventedHTTPTransport
+  from raven.transport.requests import RequestsHTTPTransport
+  from raven.transport.twisted import TwistedHTTPTransport
+  from raven.transport.tornado import TornadoHTTPTransport
+  from raven.transport.threaded_requests import ThreadedRequestsHTTPTransport
+  from raven.base import Client
+
+  for transport in [EventletHTTPTransport,
```
dcramer: of note we dont actually need to do this since you can pass transport to the Client() itself. It was purely for backwards compatibility for a period of time.

conftest.py
```diff
@@ -82,3 +90,13 @@ def pytest_configure(config):
             }],
             ALLOWED_HOSTS=['*'],
         )
+
+
+
+for transport in [EventletHTTPTransport,
```
dcramer: assuming things break without this?

gjcarneiro: Sadly yes.  The reason is that all tests build clients based on DSN, so a few test cases fail.  And I guess using the DSN and registry is part of the things we test, so we can't remove it?

```
_________________________________________________________________________________________________ ThreadedTransportTest.test_does_send __________________________________________________________________________________________________
InvalidDsn: Unsupported Sentry DSN scheme: threaded+requests+http (threaded+requests+http://some_username:some_password@localhost:8143/1)
__________________________________________________________________________________________ ThreadedTransportTest.test_shutdown_waits_for_send ___________________________________________________________________________________________
InvalidDsn: Unsupported Sentry DSN scheme: threaded+requests+http (threaded+requests+http://some_username:some_password@localhost:8143/1)
_________________________________________________________________________________________________ RequestsTransportTest.test_does_send __________________________________________________________________________________________________
InvalidDsn: Unsupported Sentry DSN scheme: requests+http (requests+http://some_username:some_password@localhost:8143/1)
________________________________________________________________________________ TornadoTransportTests.test__sending_successfully_calls_success_callback ________________________________________________________________________________
InvalidDsn: Unsupported Sentry DSN scheme: tornado+http (tornado+http://uver:pass@localhost:46754/1)
__________________________________________________________________________________ TornadoTransportTests.test__sending_with_error_calls_error_callback __________________________________________________________________________________
InvalidDsn: Unsupported Sentry DSN scheme: tornado+http (tornado+http://uver:pass@localhost:46754/1)
____________________________________________________________________________________________________ TornadoTransportTests.test_send ____________________________________________________________________________________________________
InvalidDsn: Unsupported Sentry DSN scheme: tornado+https (tornado+https://user:pass@host:1234/1?timeout=1&verify_ssl=1&ca_certs=/some/path/somefile)
```

CHANGES
gjcarneiro: But the problem is that a typical Django setup just adds `raven.contrib.django.raven_compat` to the list of INSTALLED_APPS.  If they also have a weird DSN configured, they would need to register the transports in the registry for it to work, as the Django Sentry app creates its own Client() and I honestly have no idea how to customize creation of this Client()...

dcramer: You can actually just pass it into SENTRY_CONFIG via 'transport' (same as all of the other config options).

conftest.py
dcramer: Yeah we'd have to update these tests to pass it in directly, but I'd be ok with that in order to remove the legacy compatibility.

CHANGES
gjcarneiro: Ah, OK then.  I guess I can update the documentation to reflect this.  Seems cleaner the way you suggest.

conftest.py
gjcarneiro: OK.  Will probably have to wait a few days before I commit more changes, as I'll be travelling today (to PyCon UK) and I don't have a laptop.

gjcarneiro:
Please check if the latest commits address your comments.

dcramer:
Looks good we just need to update the tornado tests to remove the tornado+ prefix I think?

gjcarneiro:
Hm, some Tornado test failing on Python 2.6, but I don't know why :(

mitsuhiko:
2.6 is very much alive, we can't kill it yet.

gjcarneiro:
It's all very fine to say "we must support Python2.6", but you have to realise it is incredibly hard, today, to do anything with Python2.6.  After much insistence, I managed to get a docker image with python2.6 and pip, but even then I didn't manage to run the tests.  The latest error is:
```
Collecting billiard<3.6.0,>=3.5.0.2 (from celery>=2.5->raven)
  Downloading billiard-3.5.0.2.tar.gz (157kB)
    100% |################################| 163kB 5.7MB/s 
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-QHA8oe/billiard/setup.py", line 90, in <module>
        raise ValueError('Versions of Python before 2.7 are not supported')
    ValueError: Versions of Python before 2.7 are not supported
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-QHA8oe/billiard/
make: *** [bootstrap] Error 1
root@45ac8d430fc8:/code# 
```
Please, please, reconsider Python 2.6, it is dragging raven-python down!

Maybe rebrand raven-python into a new major release and state in the release notes that whoever needs Python 2.6 support should continue using the old raven version?

Oh, drat, I see now you had a major release recently :(

There must be something we can do, this is ridiculous.  How are we supposed to fix tests if can't even run them?

PR #859, opened 2016-09-14
_log_failed_submission fails with malformed exceptions

bsergean:
This will manifest in those kind of errors:

```
        if 'exception' in data and 'stacktrace' in data['exception']['values'][0]:
            # try to reconstruct a reasonable version of the exception
            for frame in data['exception']['values'][0]['stacktrace']['frames']:
                output.append('  File "%(fn)s", line %(lineno)s, in %(func)s' % {
>                   'fn': frame['filename'],
                    'lineno': frame['lineno'],
                    'func': frame['function'],
                })
E               KeyError: 'filename'
```

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/859)

### DIFF
--- a/raven/base.py
+++ b/raven/base.py
@@ -657,9 +657,9 @@ class Client(object):
             # try to reconstruct a reasonable version of the exception
             for frame in data['exception']['values'][0]['stacktrace']['frames']:
                 output.append('  File "%(fn)s", line %(lineno)s, in %(func)s' % {
-                    'fn': frame['filename'],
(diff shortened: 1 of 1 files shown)


dcramer:
This seems fine since our goal should be defensiveness on this kind of entry point. Thanks!

bsergean:
Thanks ! I hope that the -1 as a default for line number won't cause problems. If it does maybe 0 can work too, but -1 is a classic "invalid" int which is why I proposed it.

PR #860, opened 2016-09-14
Add support for Django 1.10-style middleware via `MiddlewareMixin`

dtran320:
- [x] Implement `MiddlewareMixin` if we're able to import it (>= 1.10)
- [x] Update docs to mention both `MIDDLEWARE` and `MIDDLEWARE_CLASSES`
- [ ] Add tests with both old-style and 1.10-style middleware in `tests/contrib/django/tests.py`
- [ ] Configure `install_middleware()` to work with both `MIDDLEWARE` and `MIDDLEWARE_CLASSES`

This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/860)

### DIFF
--- a/docs/integrations/django.rst
+++ b/docs/integrations/django.rst
@@ -144,11 +144,11 @@ In certain conditions you may wish to log 404 events to the Sentry server. To
 do this, you simply need to enable a Django middleware:
 
 .. sourcecode:: python
-
-    MIDDLEWARE_CLASSES = (
+    # Use ``MIDDLEWARE_CLASSES`` prior to Django 1.10
+    MIDDLEWARE = (
         'raven.contrib.django.raven_compat.middleware.Sentry404CatchMiddleware',
         ...,
-    ) + MIDDLEWARE_CLASSES
+    ) + MIDDLEWARE
 
 It is recommended to put the middleware at the top, so that any only 404s
 that bubbled all the way up get logged. Certain middlewares (e.g. flatpages)
@@ -175,8 +175,8 @@ Sentry supports sending a message ID to your clients so that they can be
 tracked easily by your development team. There are two ways to access this
 information, the first is via the ``X-Sentry-ID`` HTTP response header.
 Adding this is as simple as appending a middleware to your stack::
-
-    MIDDLEWARE_CLASSES = MIDDLEWARE_CLASSES + (
+    # Use ``MIDDLEWARE_CLASSES`` prior to Django 1.10
+    MIDDLEWARE = MIDDLEWARE + (
       # We recommend putting this as high in the chain as possible
       'raven.contrib.django.raven_compat.middleware.SentryResponseErrorIdMiddleware',
       ...,
--- a/raven/contrib/django/middleware/__init__.py
+++ b/raven/contrib/django/middleware/__init__.py
@@ -14,6 +14,14 @@ import threading
 from django.conf import settings
 from django.core.signals import request_finished
 
+try:
+    # Django >= 1.10
+    from django.utils.deprecation import MiddlewareMixin
+except ImportError:
+    # Not required for Django <= 1.9, see:
+    # https://docs.djangoproject.com/en/1.10/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware
+    MiddlewareMixin = object
+
 from raven.contrib.django.resolver import RouteResolver
 
 
@@ -27,7 +35,7 @@ def is_ignorable_404(uri):
     )
 
 
-class Sentry404CatchMiddleware(object):
+class Sentry404CatchMiddleware(MiddlewareMixin):
     def process_response(self, request, response):
         from raven.contrib.django.models import client
 
@@ -52,7 +60,7 @@ class Sentry404CatchMiddleware(object):
     # sentry_exception_handler(sender=Sentry404CatchMiddleware, request=request)
 
 
-class SentryResponseErrorIdMiddleware(object):
+class SentryResponseErrorIdMiddleware(MiddlewareMixin):
     """
     Appends the X-Sentry-ID response header for referencing a message within
     the Sentry datastore.
--- a/raven/contrib/django/middleware/wsgi.py
+++ b/raven/contrib/django/middleware/wsgi.py
@@ -7,11 +7,19 @@ raven.contrib.django.middleware.wsgi
 """
 from __future__ import absolute_import
 
+try:
+    # Django >= 1.10
+    from django.utils.deprecation import MiddlewareMixin
+except ImportError:
+    # Not required for Django <= 1.9, see:
+    # https://docs.djangoproject.com/en/1.10/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware
+    MiddlewareMixin = object
+
 from raven.middleware import Sentry
 from raven.utils import memoize
 
 
-class Sentry(Sentry):
+class Sentry(Sentry, MiddlewareMixin):
     """
     Identical to the default WSGI middleware except that
     the client comes dynamically via ``get_client
--- a/tests/contrib/django/middleware.py
+++ b/tests/contrib/django/middleware.py
@@ -1,13 +1,22 @@
-class BrokenRequestMiddleware(object):
+try:
+    # Django >= 1.10
+    from django.utils.deprecation import MiddlewareMixin
+except ImportError:
+    # Not required for Django <= 1.9, see:
+    # https://docs.djangoproject.com/en/1.10/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware
+    MiddlewareMixin = object
+
+
+class BrokenRequestMiddleware(MiddlewareMixin):
     def process_request(self, request):
         raise ImportError('request')
 
 
-class BrokenResponseMiddleware(object):
+class BrokenResponseMiddleware(MiddlewareMixin):
     def process_response(self, request, response):
         raise ImportError('response')
 
 
-class BrokenViewMiddleware(object):
+class BrokenViewMiddleware(MiddlewareMixin):
     def process_view(self, request, func, args, kwargs):
         raise ImportError('view')


dtran320:
I added some TODOs above— let me know which are worth doing or if there's anything else I missed.

dcramer:
The test would def be valuable.

Are we going to need to have new style middleware in place for 1.11, or will the mixin still be available?

dtran320:
@dcramer Great question— just did some digging.

Based on this [sentence in the 2.0 deprecation release notes](https://docs.djangoproject.com/en/dev/internals/deprecation/), I'd assume `MiddlewareMixin` will stick around until 2.0:

> Support for old-style middleware using settings.MIDDLEWARE_CLASSES will be removed.

In [DEP 0005: Improved middleware](https://github.com/django/deps/blob/master/final/0005-improved-middleware.rst#transition-assistance-mixin), it says:

> The fallback from a missing MIDDLEWARE setting to MIDDLEWARE_CLASSES will be subject to a normal deprecation path. At the conclusion of that deprecation path, support for the fallback and the old-style middleware implementations in Django will be removed. The compatibility mixin will begin its own deprecation path at that point. (This allows warning-free use of the compatibility mixin throughout the deprecation path for MIDDLEWARE_CLASSES).

dcramer:
I suppose this is more relevant with Django throwing warnings at our auto injected middleware #860

jhancia:
Any ETA on these changes? Waiting to use this in production with Django 1.10. Thanks!

dcramer:
I'm going to go ahead and merge it, as theres at least basic existing coverage.

That said, I'm pretty sure this wasn't blocking usage of 1.10, it just shows warnings.

fengsi:
@dcramer Ahh just saw this.  https://github.com/getsentry/raven-python/pull/883 has the warning message fixed in `install_middleware()`!

lorddaedra:
Current version of Sentry doesn't work with Django 1.10 after some changes... and it was really hard to detect because of your ugly install_middleware(). It's bad way. Good way - tell in docs in getting started about "hey, add this to your MIDDLEWARE(_CLASSES): ... "

To detect this bug, I added ``` warnings.warn("{} {}".format(middleware_path, middleware), RemovedInDjango20Warning) ``` before ``` mw_instance = middleware(handler)``` in .../django/core/handlers/base.py

and it was your middleware. I even did not think about you installed anything... 

My error is 

```
Unhandled exception in thread started by <function check_errors.<locals>.wrapper at 0x7fef76890400>
  File "/usr/local/lib/python3.5/site-packages/django/utils/autoreload.py", line 226, in wrapper
    fn(*args, **kwargs)
  File "/usr/local/lib/python3.5/site-packages/django/core/management/commands/runserver.py", line 142, in inner_run
    handler = self.get_handler(*args, **options)
  File "/usr/local/lib/python3.5/site-packages/django/contrib/staticfiles/management/commands/runserver.py", line 27, in get_handler
    handler = super(Command, self).get_handler(*args, **options)
  File "/usr/local/lib/python3.5/site-packages/django/core/management/commands/runserver.py", line 64, in get_handler
    return get_internal_wsgi_application()
  File "/usr/local/lib/python3.5/site-packages/django/core/servers/basehttp.py", line 49, in get_internal_wsgi_application
    return import_string(app_path)
  File "/usr/local/lib/python3.5/site-packages/django/utils/module_loading.py", line 20, in import_string
    module = import_module(module_path)
  File "/usr/local/lib/python3.5/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 986, in _gcd_import
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "/usr/src/app/byvshie/wsgi.py", line 15, in <module>
    application = get_wsgi_application()
  File "/usr/local/lib/python3.5/site-packages/django/core/wsgi.py", line 14, in get_wsgi_application
    return WSGIHandler()
  File "/usr/local/lib/python3.5/site-packages/django/core/handlers/wsgi.py", line 153, in __init__
    self.load_middleware()
  File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 83, in load_middleware
    mw_instance = middleware(handler)
TypeError: Initialization arguments are not supported
```

Second line in Python Zen:
Explicit is better than implicit.

Just tell users to add this middlewares, do not patch anything without at least notes in docs (and BIG font)..

dcramer:
I can appreciate opinions but I promise you we make decisions like that carefully. The more you ask someone to do, the less often things are done correctly. 

That said it's not ok that things aren't working and we'll get this fixed asap.

lorddaedra:
Sorry, my opinion may injure someone.. But... Are you PHP-programmers?.. You do it in Python world. Seriously?.. 

If you are PHP-programmer and make Wordpress plugin, your relationships between you and user as between "blonde girl and programmer". Your user understand nothing in your area, and if you ask something, you will see beautiful empty eyes and nothing more...

But it's Django. It's Python. It's third-party app for framework, not plugin for Wordpress or other PHP CMS's. Your end users, Django programmers, have a lot of knowledge in programming. They are programmers like you, not blonde girls [sorry, girls, for stereotypes, it's just as example, I know well skilled in C++ blonde programmer...]. 

And if user install something in framework - he read installation docs and configuration options page in you docs, if upgrade - changelog and upgrade manual... Users do not asks for 2 click's solution "press to make all things good", they going to install and configure your app step by step...

This is default behaviour. People have 30-50+ apps and make their settings.py, configure MIDDLEWARE etc. (order is matter), they are not happy if all this apps will try monkey-patch ( https://ru.wikipedia.org/wiki/Monkey_patch ) their settings.py. (it's ugly practice for third-party app programmers) And no one (except may be you) try to monkeypatch MIDDLEWARE(_CLASSES) at least by default (without any mentions in docs about "please add 'OurCoolVendorMiddleware' to MIDDLEWARE(_CLASSES)")...

So first mistake is try do it by default. You can fix it by adding some lines to your docs. https://docs.sentry.io/clients/python/integrations/django/ 

Second mistake is using install_middleware() if user forget to follow installation guide (not added to MIDDLEWARE(_CLASSES) yoru middleware). You try fix it for user. No. Bad idea for Python apps. It's not PHP. Do not try to guess, just break it. Raise ImproperlyConfigured  exception or something similar. https://github.com/django/django/blob/master/django/core/exceptions.py#L87

MightySCollins:
@lorddaedra As primarily a PHP developer (we do other things apart from wordpress) I don't think you should assume we are all incompetent...

However I completely agree even in Laravel (PHP's version of Django) packages quite a few require you to add there middleware, (the ones which don't always seem to break after framework updates).

This is the behaviour in quite a few other things and if people don't follow instructions then you should polity point them in the direction of the docs. Yes it might mean more pointless GitHub issues but I think that better than magic which developers don't expect your package to do. 

I agree with lorddaedra and think that a well named and clear exception will be simple enough for most users to fix the problems themselves.

dcramer:
Not sure if you mean being a PHP developer as an insult but it has no place here.

The exception thrown here is our fault by nature but the exception handling is not. We don't control what Django does entirely when something fails, and while the issue was ultimately our problem, the obscurity in the error is unrelated to the SDK.

lorddaedra:
> Not sure if you mean being a PHP developer as an insult but it has no place here.

It's just another use case. If you make code for end users (as most PHP programmers do) - by default you make decisions for user. (Any CMS: ModX, Wordpress, Joomla, etc.)

If you make code for frameworks users (Laravel (PHP) / Symfony (PHP) / Ruby on Rails (Ruby) / Django (Python) / Play 2 (Java/Scala) etc. ) - they prefer control all things. (And here in Django this type of magic  for this type of tasks is antipattern). In RoR or some PHP-frameworks it may be more permissive ( http://stackoverflow.com/questions/441717/whats-wrong-with-magic ), but in Django we do not like magic, we prefer to control everything and make decisions.. [ If it's not about magic pony :-) http://www.djangopony.com/ ]

Again, third-party apps for Django should not patch anything in my project files, config settings, etc. at least without BIG warning (and even with warnings about magic it looks wrong). MIDDLEWARE should be controlled by project programmer, not by third-party app programmer.

hdra:
@lorddaedra You can make your point without having to put out all that passive-aggressive statements. That doesn't really make it any easier for anyone to fix this either.

That said, personally I agree with having something more explicit, or at least put something that explains what installing `raven-python` does to django, exactly, in the docs (maybe something clearer than "install a hook in Django"?). Though I'm not sure how to go about this without having the docs getting too verbose explaining all the edge cases.

lorddaedra:
> You can make your point without having to put out all that passive-aggressive statements. That doesn't really make it any easier for anyone to fix this either.

May be I was too emotional. I did not expect from Sentry.. I hope no one is offended...

About docs: good idea look to other apps, for example, django debug toolbar: https://django-debug-toolbar.readthedocs.io/en/stable/installation.html
Here they tell user how to install and configure their app: add to INSTALLED_APPS some apps, improve URLconf, add middleware, add some configuration variable (optional).

lorddaedra:
(And you can find much more examples of Django apps [and their docs] here: https://djangopackages.org/ , I think, 99% of them will ask user to manually add middleware if app require this step.)

mattrobenolt:
> May be I was too emotional. I did not expect from Sentry.. I hope no one is offended...

I'm just going to jump in here and curb this now. Comments you've made here are absolutely offensive to some and will not be tolerated here. We're open for healthy discussions and disagreements, but by insulting females, PHP developers, and frankly, us, is not the way to do it.  Please don't do this again.

dtran320:
Sorry, have been super busy at work and haven't checked back in on this— I'm sad to see some vitriol in a PR that I opened.

@mattrobenolt @hdra @dcramer: Really love and appreciate your professionalism in handling this.

@lorddaedra: Thanks for your thoughts, although I hope you can keep things positive from here on out.

What introduced the breaking change for @lorddaedra? Was it this changeset or potentially https://github.com/getsentry/raven-python/pull/883/files? @lorddaedra I can't reproduce that error— did this happen with Django 1.10— which version of `raven` do you have installed? It's not clear when you say "current" if you mean the latest release version or development version.

And what would be the best way forward? Would asking the user to add `raven.contrib.django.middleware.SentryMiddleware` and `raven.contrib.django.middleware.SentryLogMiddleware`to `MIDDLEWARE` or `MIDDLEWARE_CLASSES` and changing `install_middleware` to something like `check_middleware` which raises an exception if these aren't added?

dtran320:
Looks like this changeset might fix a remaining instance where I forgot `MiddlewareMixin`: https://github.com/getsentry/raven-python/issues/886

dtran320:
Nevermind— looks like @mitsuhiko is on top of things. Everything works for me in `5.31.0`

lorddaedra:
> I'm just going to jump in here and curb this now. Comments you've made here are absolutely offensive to some and will not be tolerated here. We're open for healthy discussions and disagreements, but by insulting females, PHP developers, and frankly, us, is not the way to do it. Please don't do this again.

Did I insulted Django pony too? (I said, I do not like magic) 
I insult only bad design of app ( I mean only install_middleware() here ) and lack of information in docs (Installation page), nothing more. In my opinion using this type of magic for installation apps in Django just unprofessional (do not read it as personal insult but as bug in app code/design) and this type of magic acceptable only for making PHP plugins for CMS. (Where is  insulting PHP developers here?.. they just work for another use cases scenario... and if you create Sentry PHP plugins for CMS - you can do it, why not.. but for Django it's antipattern and I was little surprised about you wanna discuss it. You can just open any other installation manuals and see that noone in Django world uses such methods [for example, manual of closest Sentry alternative, https://rollbar.com/docs/notifier/pyrollbar/#django , they ask users to add some settings too] )


> What introduced the breaking change for you, @lorddaedra? 
Release raven==5.30.0. ( SentryMiddleware )

Can't test it just now, but can in next 2-3 days, I will post here if will find something..

dcramer:
@lorddaedra you can stop trying to educate us on Django. We've been working and contributing to it since version 0.95. I will say with confidence that our team has a very strong understanding of both Django as well as the Python ecosystem. 

While I don't mind **constructive** conversation around how we're doing things, at the end of the day we control Sentry, and we make the decisions for the projects. Whether those decisions are right or wrong, they're our decisions. 

This conversation however, is not constructive, and quite frankly I'm fucking tired of it showing up in my inbox. You continue to attempt to belittle the developers thinking that poking fun of PHP has any relevance here, and more importantly you assume you know better. That will not be tolerated within our communities.

PR #861, opened 2016-09-16
Improve exception handling in get_user_info

broundee:
The original code might break, e.g. when `user` is an instance of `django.utils.functional.SimpleLazyObject` and its `_setup` raises exception.

We need to make sure the exception gets reported regardless of the user info availability.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/861)


broundee:
Will clean up the commits and create another one.

PR #862, opened 2016-09-16
Improve exception handling in get_user_info

broundee:
The original code might break on `hasattr(user, 'is_authenticated')` call, e.g. when `user` is an instance of `django.utils.functional.SimpleLazyObject` and its `_setup` raises exception.

We need to make sure the exception gets reported regardless of the user info availability.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/862)


dcramer:
Agreed, thanks!

PR #874, opened 2016-09-29
Fix unreferenced variable in get_user_info

broundee:
The previous fix (#862) left the code broken, since in case of exception, `user_info` was not declared.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-python/874)


dcramer:
Thanks!

PR #888, opened 2016-10-18
Improve exception handling in Serializer.transform

broundee:
This is an additional fix for the case described in #862.

Same happens here: when `serializer.can()` triggers evaluation of the `value` of type `django.utils.functional.SimpleLazyObject` and the latter throws exception, the error isn't reported.


broundee:
@dcramer could you please review this?

broundee:
@mitsuhiko could you please take a look at this issue when you have time? It's a really small change and has been hanging here for a while now.

PR #889, opened 2016-10-18
Change django release path

MeredithAnya:
Default path name gives error:
screen shot 2016-10-18 at 12 57 13 pm

Changed the path so that it doesn't have this error. User might still have to configure their path differently (e.g. '../..').

@benvinegar


benvinegar:
🎉

PR #893, opened 2016-10-21
use deprecation MiddlewareMixin in SentryMiddleware ; should really fix #886

### LINKED ISSUE #886: SentryMiddleware problem with Django 1.10: TypeError: Initialization arguments are not supported
After updating from `5.27.1` to `5.28.0`, I get the error:

      File ".../wsgi.py", line 10, in <module>
        application = get_wsgi_application()
      File ".../venv/lib/python3.5/site-packages/django/core/wsgi.py", line 14, in get_wsgi_application
        return WSGIHandler()
      File ".../venv/lib/python3.5/site-packages/django/core/handlers/wsgi.py", line 153, in __init__
        self.load_middleware()
      File ".../venv/lib/python3.5/site-packages/django/core/handlers/base.py", line 84, in load_middleware
        mw_instance = middleware(handler)
      File ".../venv/lib/python3.5/site-packages/gevent/local.py", line 242, in __new__
        raise TypeError("Initialization arguments are not supported")
    TypeError: Initialization arguments are not supported

The error is raised in loading `raven.contrib.django.middleware.SentryMiddleware`. It seems related to the new change to fix Django 1.10 middleware warning.

Hopefully I'll check to provide a patch. I'm using `Django==1.10.2` using the older style `MIDDLEWARE = []` settings, if that is related.

arthru:
The problem was that Middleware interface changed in Django 1.10 and SentryMiddleware did not use that new interface.

### DIFF
--- a/raven/contrib/django/middleware/__init__.py
+++ b/raven/contrib/django/middleware/__init__.py
@@ -78,7 +78,7 @@ class SentryResponseErrorIdMiddleware(MiddlewareMixin):
         return response
 
 
-class SentryMiddleware(MiddlewareMixin, threading.local):
+class SentryMiddleware(threading.local, MiddlewareMixin):
     resolver = RouteResolver()
 
     # backwards compat
--- a/raven/contrib/django/middleware/__init__.py
+++ b/raven/contrib/django/middleware/__init__.py
@@ -78,7 +78,7 @@ class SentryResponseErrorIdMiddleware(MiddlewareMixin):
         return response
 
 
-class SentryMiddleware(threading.local):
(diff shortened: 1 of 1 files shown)


arthru:
Well, obviously, this does not work with django !=1.10. I'll rework that later...

chaffeqa:
Could this be added? its currently single-handedly preventing any app from adopting the 1.10 middleware upgrade

dcramer:
@chaffeqa this behavior is not the same as whats in master and this is just a remant of a PR that slipped through (another was landed which solved this differently)

PR #901, opened 2016-10-30
Drop Python 2.6 support

di:
Yesterday marks 3 years since the last release of Python 2.6! 🎉

To celebrate, I'm attempting to drop support for it from 156 prominent Python packages (one for every week it's past end-of-life)--including this one!

I've tried my best to remove as much 2.6-specific cruft as I can, but at the very least, this PR will remove the `'Programming Language :: Python :: 2.6'` trove classifier from this projects `setup.py`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ cache:
 #  on:
 #    tags: true
 python:
-  - '2.6'
   - '2.7'
 #  - '3.2'
   - '3.3'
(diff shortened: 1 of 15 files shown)


mattrobenolt:
Thanks, but no thanks. Unless we have an absolute valid reason for dropping 2.6, we aren't. Lots of people still use 2.6 and it's not our agenda to push. Going out of our way to drop support is counter productive for a library such as ours.

PR #906, opened 2016-11-10
Improve matrix builds; Drop Python 2.6

dcramer:
- Add Flask-specific builds
- Add Celery-specific builds
- Restore Django dev build
- Reduce Django builds to only focus on subset of versions
- Expand norecursedirs (fixes plain ``py.test`` when certain artifacts are present)
- Drop formal (tested) Python 2.6 support
- Lock pytest-django based on Django version
- Lock django-celery versions
- Fix an issue affecting specific versions of Celery (missing ``throws`` on tasks)


dcramer:
Call this step one of having real test suites.

Id like to expand on this so the DJANGO envs are only running the django subtests, and the non-DJANGO envs dont run those integration tests.

Additionally it'd be nice to do similar things for Flask, and other core integrations.

dcramer:
Agenda here is to get things green and put hard locks on changing any build requirements in the future as it looks like a number of things have been broken by recent PRs.

dcramer:
Ok pretty sure I'm content here. This ended up being a bigger rabbit hole than I thought.

mitsuhiko:
I'm okay with that but I think we need to at least test 2.6 every once in a while. There are still lots of users on it.

dcramer:
@mitsuhiko we could add a suite for it but it was broken (we could also just say they have to use last major release, as its not overly hard to upgrade from 2.6 to 2.7)

mitsuhiko:
Main reason some people are stuck on 2.6 is centos and rhel.

PR #909, opened 2016-11-20
Add settings var to allow customizing sentry middleware

jsferrei:
We'd like the ability to override certain aspects of the middleware functionality without monkey-patching. Specifically, _get_transaction_from_request does not work well with some of our urls, so we want to have a fallback if nothing is resolved.

### DIFF
--- a/raven/contrib/django/models.py
+++ b/raven/contrib/django/models.py
@@ -227,7 +227,7 @@ def install_middleware():
     This ensures things like request context and transaction names are made
     available.
     """
-    name = 'raven.contrib.django.middleware.SentryMiddleware'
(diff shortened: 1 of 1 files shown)


dcramer:
@jsferrei I'd rather we provide a hook for you to add your own transaction handling vs risking people breaking the SDK by overriding the middleware with something thats not future proof. This could be as simple as moving the helper to the Django Client and letting you override it there.

jsferrei:
I'm into that. Close this?

dcramer:
:+1:

dcramer:
Ah actually I just looked at how the code was working. We'd probably want to pass it into Client, yeah.

PR #910, opened 2016-11-20
Merge logger tags

dcramer:
When using both handler defined and logger defined tags, merge them allowing the handler to override the defaults.


mitsuhiko:
We should probably put that into the changelog.

PR #911, opened 2016-11-21
Move get_transaction method to the client

jsferrei:


jsferrei:
@dcramer is this what you had in mind? Seems to function the way I'd want/expect it to.

raven/contrib/django/client.py
```diff
@@ -291,3 +293,11 @@ def capture(self, event_type, request=None, **kwargs):
             }
 
         return result
+
+    def get_transaction_from_request(self, request):
+        # TODO(dcramer): it'd be nice to pull out parameters
```
jsferrei: Hah! yes.

jsferrei:
@dcramer dynamically generated urlconfs

PR #913, opened 2016-11-21
In Django HTTP request data, multiple values for same key are lists

fdintino:
In a Django HttpRequest object, the request data is a QueryDict object, which is a dictionary-like class that supports multiple values for the same key. This is used, for instance, with `<select multiple>` form elements.

In order to have the request data serialize into JSON correctly when there are multiple values, it is necessary to use `QueryDict.iterlists()` / `QueryDict.lists()` (python 2 and 3, respectively). Otherwise, the value passed to Sentry will contain only the last value in the query string.

### DIFF
--- a/raven/contrib/django/client.py
+++ b/raven/contrib/django/client.py
@@ -16,6 +16,8 @@ from django.conf import settings
 from django.core.exceptions import SuspiciousOperation
 from django.http import HttpRequest
 from django.template import TemplateSyntaxError
+from django.utils.datastructures import MultiValueDict
+from django.utils import six
 
 try:
     # support Django 1.9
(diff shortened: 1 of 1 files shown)


raven/contrib/django/client.py
```diff
@@ -202,6 +204,15 @@ def get_data_from_request(self, request):
                         data = request.POST or '<unavailable>'
                     except Exception:
                         data = '<unavailable>'
+                    else:
+                        def list_or_item(o):
```
dcramer: can we define this at module scope to avoid computing a new function each time

fdintino: Sure.

fdintino:
The logic is simple enough that I replaced the function with a conditional expression.

fdintino:
@dcramer I've incorporated your feedback. Let me know if I need to make any other changes. Thanks!

dcramer:
lgtm but will let @mitsuhiko sign off and merge since he has more context lately

PR #914, opened 2016-11-21
Add last_event_id to Client

dcramer:


mitsuhiko:
Looks good but I swear we already had that. Am I imagining things?

dcramer:
@mitsuhiko only in Flask afaict

mitsuhiko:
I guess we should make the Flask client also set that attribute in addition to the version on `g`.

mitsuhiko:
Not here though: https://github.com/getsentry/raven-python/blob/40b368e2ec1c245141414343d7c4d51a5aca85c1/raven/contrib/flask.py#L116-L130

dcramer:
@mitsuhiko that calls out to capture which does it. We dont use the same last_event_id since the behavior here differs and resets on each request

PR #924, opened 2016-12-05
Do not record errors for terminals

mitsuhiko:
This will not record errors by default if a tty is detected.  If you
want to record anyways do not register the default hook and manually
install it with the flag flipped.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,7 @@ Version 6.0.0
 * Refactored transports to support multiple URLs.  This might affect
   you if you have custom subclasses of those.  The main change is that
   the URL parameter moved from the constructor into the `send` method.
+* Do not record from the sys except hook if stdin is a terminal.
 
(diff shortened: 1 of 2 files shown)


dcramer:
offline conversation around this was how noisy things like custom scripts are when you're fucking around from a terminal

+1

mitsuhiko:
@dcramer added test

mattrobenolt:
Why not just entirely skip installing the hook if it's not a tty? Is it possible that stdin being a tty could change while the process is running?

mitsuhiko:
@mattrobenolt i don't know if it can change on the same stream but there are a few cases where this can be an issue:

* you fork off a child and close a input stream
* someone patches sys.stdin at runtime

ryanhiebert:
We've been going nuts with typos on an interactive console causing emails from Sentry. We would really appreciate this feature!

PR #928, opened 2016-12-13
uwsgi: also allow `threads` conf setting

mattrobenolt:
uwsgi implicitly enables threads when `threads` are set without the need
to set `enable-threads`.

I've verified this behavior locally even when passing `--threads 0`.
uwsgi only needs the key to exist.

### DIFF
--- a/raven/utils/compat.py
+++ b/raven/utils/compat.py
@@ -54,6 +54,11 @@ def check_threads():
     except ImportError:
         return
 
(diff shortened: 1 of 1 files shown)


rkrzr:
We were running into this issue, but good to see that it's already fixed. Are you planning a new release soon?

mitsuhiko:
Yes. I will push one out today if there are no regressions.

PR #930, opened 2016-12-19
Support release and environment tags for Zope

fredvd:
In zope.conf's configuration spec, which will be passed on and processed by Raven.


mattrobenolt:
Thanks! 🍦

PR #932, opened 2016-12-22
Support threaded transport also for background or server processes

frietsch-ot:
We are running the raven python client inside of a (yes, very old school, I agree) CGI web application and would like to use a threaded transport, so that normal operation can continue while raven is working.

However, in rare but not rare enough cases, the CGI process is about to terminate before raven has sent all messages, and then, some output "Sentry is attempting to send... Hit Ctrl+C to abort" appears on stdout, which - lame old CGI - is the web page sent to the client. In worst case, the output is seen as (broken) HTTP headers, leading to interesting HTTP status codes from the web server.

To avoid the stdout output, I added some ThreadedBackgroundHTTPTransport, which uses the same timeout approach as the original ThreadedHTTPTransport, but simply doesn't print the Ctrl+C message. My updated test case uses both transports for complete test coverage, even though the change is rather boring.

Note: The overall diff is very small in fact, it's just that I had to change a lot of indent.

### DIFF
--- a/raven/transport/threaded.py
+++ b/raven/transport/threaded.py
@@ -26,7 +26,7 @@ logger = logging.getLogger('sentry.errors')
 class AsyncWorker(object):
     _terminator = object()
 
-    def __init__(self, shutdown_timeout=DEFAULT_TIMEOUT):
+    def __init__(self, shutdown_timeout=DEFAULT_TIMEOUT, interactive=True):
         check_threads()
         self._queue = Queue(-1)
         self._lock = threading.Lock()
@@ -34,6 +34,7 @@ class AsyncWorker(object):
         self._thread_for_pid = None
         self.options = {
             'shutdown_timeout': shutdown_timeout,
+            'interactive': interactive,
         }
         self.start()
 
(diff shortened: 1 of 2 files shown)


mattrobenolt:
Would it help if we just wrote to stderr instead of stdout? Not sure if stderr would affect CGI output or not.

I think we should also detect `isatty()` here as well since prompting for `Ctrl-C` is silly if you don't have a TTY.

frietsch-ot:
Writing to stderr doesn't seem to be an optimal choice, as this output is written to the server error log then (but I don't consider this "wanna wait" to be a real error :-) ). Certain HTTP servers may create HTTP 500s when reading anything on stderr.

But: Checking sys.stdout.isatty() is a great idea, Matt! This indeed seems to be "False" in my CGI environment (and True otherwise, of course) and should solve my problem with a very very short diff to your original code.

Sadly, I'm not yet very experienced in pull requests on GitHub (I assume it's easier to create a new one from a fresh branch instead of trying to fix the original PR with another commit). How shall we proceed?

frietsch-ot:
Ok, I'll create a separate pull request with the isatty() approach, to get things forward 😃

PR #933, opened 2016-12-29
Move get_transaction_from_request to client to allow overrides

jsferrei:
@dcramer @mitsuhiko this is a redo of https://github.com/getsentry/raven-python/pull/911
-- I went to fix conflicts and severely messed up history, so figured it better to recreate as new

### DIFF
--- a/raven/contrib/django/client.py
+++ b/raven/contrib/django/client.py
@@ -29,6 +29,7 @@ from raven.base import Client
 from raven.contrib.django.utils import get_data_from_template, get_host
 from raven.contrib.django.middleware import SentryLogMiddleware
 from raven.utils.compat import string_types, binary_type, iterlists
+from raven.contrib.django.resolver import RouteResolver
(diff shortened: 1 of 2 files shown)


mitsuhiko:
From what I can tell this looks okay. Do we want this @dcramer?

raven/contrib/django/client.py
```diff
@@ -297,3 +299,7 @@ def capture(self, event_type, request=None, **kwargs):
             }
 
         return result
+
+    def get_transaction_from_request(self, request):
+        result = self.resolver.resolve(request.path)
```
mattrobenolt: Can we skip the unnecessary variable here please?

jsferrei:
@dcramer @mitsuhiko Any update on this/requested changes? Would like this so we can override the default and leverage sentry on our app stack.

mattrobenolt:
Thanks! 🥇

jsferrei:
Thanks!

PR #939, opened 2017-01-14
Minor readability improvement

Chronial:


mattrobenolt:
You've changed behavior. `max()` is more correct.

Chronial:
I wasn't sure whether this change might be too small to submit, but you just made a perfect argument that it's not ^^.

with `timeout = 5`:
```
initial_timeout = 0.1
if 5 < 0.1:
     initial_timeout = timeout
```
-> `initial_timeout == 0.1 == min(0.1, 5)`

with `timeout = 0.01`:
```
initial_timeout = 0.1
if 0.01 < 0.1:
     initial_timeout = timeout
```
-> `initial_timeout == 0.01 == min(0.1,  0.01)`

mattrobenolt:
Derp. You're right. It's early here. :)

PR #941, opened 2017-01-17
Only print to stdout if stdout is connected to a tty(-like) device

frietsch-ot:
We are running the raven python client inside of a (yes, very old school, I agree) CGI web application and would like to use a threaded transport, so that normal operation can continue while raven is working.

However, in rare but not rare enough cases, the CGI process is about to terminate before raven has sent all messages, and then, some output "Sentry is attempting to send... Hit Ctrl+C to abort" appears on stdout, which - lame old CGI - is the web page sent to the client. In worst case, the output is seen as (broken) HTTP headers, leading to interesting HTTP status codes from the web server.

To avoid the stdout output, as discussed in https://github.com/getsentry/raven-python/pull/932, I just added an isatty() check.

So: The overall diff is very small in fact, it's just that I had to change a bit of indent.

### DIFF
--- a/raven/transport/threaded.py
+++ b/raven/transport/threaded.py
@@ -11,6 +11,7 @@ import atexit
 import logging
 import threading
 import os
+import sys
 
 from time import sleep, time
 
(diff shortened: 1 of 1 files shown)


raven/transport/threaded.py
```diff
@@ -64,18 +65,20 @@ def main_thread_terminated(self):
...
-                print("Sentry is attempting to send %i pending error messages"
-                      % size)
-                print("Waiting up to %s seconds" % timeout)
-
-                if os.name == 'nt':
-                    print("Press Ctrl-Break to quit")
-                else:
-                    print("Press Ctrl-C to quit")
+
+                if sys.stdout.isatty():
```
mattrobenolt: Should test against stdin, not stdout.

frietsch-ot: Hm... as we are printing to stdout, I'd propose to - if you want to test for stdin - to test for both. At least, I also don't want to see those messages when redirecting a console application to a file, maybe by nohup or something.

>>> import sys
>>> sys.stdout.isatty()
True

on my machine.

frietsch-ot:
How do you think about that "if sys.stdin.isatty() and sys.stdout.isatty()" in my last commit?

Maelcom:
Would be great to see this merged anytime soon.

katyukha:
Hi all,

We have similar issue, but i think it would be better to print these messages to STDERR instead of STDOUT.

Issue #957, opened 2017-02-15
Request is not logged to sentry using the default django middleware


teeberg:
It looks like there's code to automatically track requests and send them to sentry even when you don't pass `extra={'request': request}` to the logging call. However, it's not working for us because of the following:

The request gets set in local thread storage [here](https://github.com/getsentry/raven-python/blob/6ead8beb09d6d40b91be213580dd4d9d5bd72f37/raven/contrib/django/middleware/__init__.py#L105), where `self.thread` is an instance of `raven.contrib.django.middleware.SentryMiddleware`, and then read [here](https://github.com/getsentry/raven-python/blob/6ead8beb09d6d40b91be213580dd4d9d5bd72f37/raven/contrib/django/client.py#L258), where `SentryLogMiddleware.thread` is a `property` object, which is not what it was set on, and thus leads to `request` being `None`.

Doing some git blaming, I think this broke in https://github.com/getsentry/raven-python/commit/c3e504a0bf77deb9a79bd841e9cacc62c8127901, where `SentryMiddleware.thread` was changed from a thread-local class variable to a property

This is with raven-python 5.32.0

dcramer:
Not sure how we fucked this one up, but I'll dig into it as glancing over you're definitely right.

teeberg:
That was crazy fast! Thank you for the really quick fix!! :)

Issue #940, opened 2017-01-16
Resolver returns a bad url


koleror:
I'm using Django 1.7 (I know) and I just upgraded from raven 5.24.3 to 5.32.0, which causes me some new issues with the new UrlResolver. In some case, the culprit get translated from path like `v1/cart/mine/items/` to `/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{api_name}/{resource_name}/{pk}/items` which then goes to the transaction tag in sentry.

I investigated a bit and found out the culprit is `raven.contrib.django.resolver.RouteResolver._resolve` which calls itself recursively and increase the `prefix` size for whatever reason...  
Do you have any clue how I can fix this please?

dcramer:
@koleror could you provide some details on how the route is formed? that would help us write a test case to ensure we can resolve the behavior

koleror:
I'm not sure what you need exactly?

EdgyEdgemond:
I too have this issue.

I believe it is to do with the base (maybe also child urls.py files) having multiple entries...

 14     url(r'^api/(?P<version>v[\d\.]+)/', include('apps.api.urls.v1.one')),
 15     url(r'^api/(?P<version>v[\d\.]+)/', include('apps.api.urls.v1.two')),
 16     url(r'^api/(?P<version>v[\d\.]+)/', include('apps.api.urls.v1.three')),
 17     url(r'^api/(?P<version>v[\d\.]+)/', include('apps.api.urls.v1.four')),

This means if it matches v1.one you get 'api/{version}/one'

If it matches v1.two you get 'api/{version}/api/{version}/two'

etc etc.

koleror:
Hmm could be, as I'm using multiple includes as well
Nice to know I'm not the only one out there :D

koleror:
In case it may help, the buggy routes are generated through the [Django Tastypie](https://github.com/django-tastypie) plugin

Thanks for your help

dvndrsn:
We're also facing a similar issue on Tastypie 0.12.1 with Django 1.7.9. I did some analysis on this (below).

[...]

PR #958, opened 2017-02-15
[django] fix shared request not being available in thread context

dcramer:
Fixes https://github.com/getsentry/raven-python/issues/957


teeberg:
Is there a regular release schedule that you guys follow? When is this expected to go to PyPI?

dcramer:
I was hoping to resolve https://github.com/getsentry/raven-python/issues/940 before we pushed out a next release, but so far haven't been able to. We'll either get one out today or early next week.

teeberg:
Oh yeah, we've also been seeing #940 for a while :-) Maybe I can figure out what's causing it

PR #959, opened 2017-02-15
[django] expand tests to cover tastypie

dcramer:
This also corrects an issue with recursion behavior in transaction names, exhibited by standard tastypie URLs.

Fixes GH-940

### DIFF
--- a/raven/contrib/django/resolver.py
+++ b/raven/contrib/django/resolver.py
@@ -56,8 +56,8 @@ class RouteResolver(object):
 
         if parents is None:
             parents = [resolver]
-        else:
-            parents.append(resolver)
+        elif resolver not in parents:
+            parents = parents + [resolver]
 
         new_path = path[match.end():]
         for pattern in resolver.url_patterns:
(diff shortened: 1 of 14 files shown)


dcramer:
breaking django changes out into https://github.com/getsentry/raven-python/pull/960 since I have no idea why the tests dont pass on Travis

dcramer:
@mitsuhiko @mattrobenolt one of you mind taking a quick pass on review here?

raven/contrib/django/resolver.py
```diff
@@ -56,8 +56,8 @@ def _resolve(self, resolver, path, parents=None):
 
         if parents is None:
             parents = [resolver]
-        else:
-            parents.append(resolver)
+        elif resolver not in parents:
+            parents = parents + [resolver]
```
mitsuhiko: is this here guaranteed to be a list?

dcramer: @mitsuhiko yeah parents is only set by this function

mitsuhiko: I thought parents is passed in?

dcramer: only by itself (recursively), and only as a list

PR #962, opened 2017-02-17
Allow subclasses of SentryHandler to be modified by the celery logging signal, resolves #961

### LINKED ISSUE #961: raven.contrib.celery.register_logger_signal ignores subclasses of SentryHandler
`raven.contrib.celery.register_logger_signal` attempts to find an existing handler to add a CeleryFilter to.  It only considers existing handlers that are of type: `raven.handlers.logging.SentryHandler`.  This means that if a subclass of this handler is found (such as `raven.contrib.django.handlers.SentryHandler`) the existing handler is ignored and a brand new handler with default settings is added, resulting in two potentially duplicate handlers.

The fix for this would be to change the `type` check [here](https://github.com/getsentry/raven-python/blob/master/raven/contrib/celery/__init__.py#L46) to an `isinstance` check.

I'm interested in submitting a PR for this with tests, if this change is considered appropriate.

Here is the skeleton of a test that would help cover this:
```
def test_subclassed_handler_updated(self):
    client = InMemoryClient()
    register_logger_signal(client)

    class CustomHandler(SentryHandler):
        pass

    root = logging.getLogger()
    handler = CustomHandler(client)
    root.addHandler(handler)

    celery.signals.after_setup_logger.send(
        sender=None, logger=root,
        loglevel=logging.WARNING, logfile=None,
        format=u'', colorize=False,
    )

    assert len(root.handlers) == 1

    handler = root.handlers[0]
    assert isinstance(handler, CustomHandler)

    assert len(handler.filters) == 1

    _filter = handler.filters[0]
    assert isinstance(_filter, CeleryFilter)
```

mpistrang:
This change will allow the celery logging signal to recognize subclasses of `raven.handlers.logging.SentryHandler` when searching for existing handlers to add a CeleryFilter to, instead of adding a second handler.  This resolves #961

### DIFF
--- a/raven/contrib/celery/__init__.py
+++ b/raven/contrib/celery/__init__.py
@@ -43,7 +43,7 @@ def register_logger_signal(client, logger=None, loglevel=logging.ERROR):
         # that the CeleryFilter is installed.
         # If one is found, we do not attempt to install another one.
         for h in logger.handlers:
-            if type(h) == SentryHandler:
+            if isinstance(h, SentryHandler):
                 h.addFilter(filter_)
                 return False
 
--- a/tests/contrib/test_celery.py
+++ b/tests/contrib/test_celery.py
@@ -1,8 +1,14 @@
 from __future__ import absolute_import
 
 import celery
+import logging
 
-from raven.contrib.celery import SentryCeleryHandler
+from raven.contrib.celery import (
+    SentryCeleryHandler,
+    register_logger_signal,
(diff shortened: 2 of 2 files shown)


mpistrang:
Thanks @ashwoods for the approval.  Do you need anything else from me for this to be merged?

ashwoods:
Just wanted to manually rebase it and run the tests again locally. will merge tomorrow morning.

PR #964, opened 2017-02-17
Handle ignore_exceptions values being class or string

### LINKED ISSUE #963: Ignoring exceptions on Flask does not work anymore with child exceptions
#821 is a fundamental behavior change on change on exception handling which is not documented into the changelog.
The main consequence is that child exception are now reported to sentry unless you list all child exceptions.
On a big project, this is really painful and this ends with a sentry massive polution with unwanted expection.

On flask, as handled HTTP errors are exception-based, I have to switch from ignoring HTTPException to ignoring every child, the werkzeug ones and the app children ones. In a case where the app can also host plugins raising custom exception, it becomes a nightmare.

noirbizarre:
This PR handle `ignore_exceptions` values being either string or class.
In the case of class, child exceptions are ignored too.

This both fixes #963 bringing back the existing flask behavior and allows all other integrations to take advantage of this behavior.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,8 @@ Version 6.0.0
   affected things like django-tastyepie.
 * Corrected an issue where Django's HTTP request was not always available
   within events.
+* Support both string and class values for ``ignore_exceptions`` parameters.
+  Class values also support child exceptions.
 
 Version 5.32.0
 --------------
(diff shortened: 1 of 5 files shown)


noirbizarre:
@dcramer any chance to have this one reviewed so I have time ⌛️ to fix if necessary before the next release ?

noirbizarre:
Thanks !!! 🎉 🎊

Issue #884, opened 2016-10-17
patch_cli_runner was removed and it appears that raven no longer automatically sends django command errors


budlight:

mitsuhiko:
The default sys.excepthook should capture it. With which version of Django do you see it?

currankaushik:
I don't have context for the original issue, but I think I may be encountering a similar problem, where exceptions in my management commands are not being captured by Raven. 

As an example, I created a dummy command that just raises an exception in its `handle` method. In Raven versions <5.24.0, running the command via `./manage.py mycommand` will present the exception message on the console, along with a message along the lines of `Sentry is attempting to send 2 pending error messages`, and the exception will make it to a dummy Sentry project I've set up. In Raven versions >=5.24.0, running the command simply dumps the exception message to the console, and the exception does not make it to the Sentry project.

I've tested this on Django 1.10.3 and 1.9.9 with the same results.

My Raven and logging configs are the following:
```
RAVEN_CONFIG = {
  "dsn": "<mydsn>",
}
LOGGING = {
  "version": 1,
  "disable_existing_loggers": False,
  "root": {
    "level": "INFO",
    "handlers": ["sentry"],
  },
  "handlers": {
    "sentry": {
      "level": "WARNING",
      "class": "raven.contrib.django.raven_compat.handlers.SentryHandler",
    }
  }
}
```

In addition, I have a custom client class (basically just a subclass of the `DjangoClient` that passes `install_sql_hook=False`). I have included it in my Django `settings.py` file as so:
```
SENTRY_CLIENT = "path.to.my.client"
```

I've tried to investigate this issue by seeing how the custom excepthook is set up in Raven [here](https://github.com/getsentry/raven-python/blob/master/raven/base.py). It appears that `sys.excepthook` is only set when the `Client` is instantiated, but it doesn't seem like the `Client` is getting instantiated when the management command is run.

Please let me know if any additional information would be helpful.

mitsuhiko:
Can you paste the actual traceback you get on the console? I wonder if Django now catches those exceptions itself and reports them.

currankaushik:
```
Traceback (most recent call last):
  File "./manage.py", line 13, in <module>
    execute_from_command_line(sys.argv)
  File "/home/ck/env/lib/python2.7/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
[...]

PR #966, opened 2017-02-21
[django] fix registration of hooks

dcramer:
- utilize app.ready() for Django 1.7+
- ensure client is instantiated upon initialization (fixes sys.except_hook)

Fixes GH-884

### DIFF
--- a/raven/contrib/django/models.py
+++ b/raven/contrib/django/models.py
@@ -237,13 +237,14 @@ def install_middleware():
                     type(middleware)((name,)) + middleware)
 
 
-if (
-    'raven.contrib.django' in settings.INSTALLED_APPS or
-    'raven.contrib.django.raven_compat' in settings.INSTALLED_APPS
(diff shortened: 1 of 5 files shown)


raven/contrib/django/models.py
```diff
@@ -237,13 +237,14 @@ def install_middleware():
                     type(middleware)((name,)) + middleware)
 
 
-if (
-    'raven.contrib.django' in settings.INSTALLED_APPS or
-    'raven.contrib.django.raven_compat' in settings.INSTALLED_APPS
-):
+def initialize():
```
mitsuhiko: Maybe this should make sure it only runs once? I am worried this wraps the WSGI and other things twice if run twice.

dcramer: added global lock here + initializer check

mitsuhiko:
jftr. only read comment :P

PR #970, opened 2017-02-28
Always supply a user.ip_address value

mattrobenolt:
When applicable, we should try and pass along the X-Forwarded-For
value from the client as REMOTE_ADDR. There is no security implications
here and only yields more accurate data in Sentry itself. It's not as
common for a user to mutate their own REMOTE_ADDR, so we should try to
do the right thing and fetch XFF when present instead.

It's an explicitly decision to mutate REMOTE_ADDR inside of the `env`
rather than the `user` interface because right now, the Sentry server is
taking the liberty of filling in the user.ip_address from the
env.REMOTE_ADDR already when it exists, so it only makes sense to just
pass along the more expected behavior.

This has come up a number of times in support and from users.

### DIFF
--- a/tests/contrib/flask/tests.py
+++ b/tests/contrib/flask/tests.py
@@ -313,7 +313,7 @@ class FlaskLoginTest(BaseTest):
 
     def test_user(self):
         self.client.get('/login/')
-        self.client.get('/an-error/')
+        self.client.get('/an-error/', environ_overrides={'REMOTE_ADDR': '127.0.0.1'})
         event = self.raven.events.pop(0)
         assert event['message'] == 'ValueError: hello world'
         assert 'request' in event
--- a/tests/contrib/flask/tests.py
+++ b/tests/contrib/flask/tests.py
@@ -59,7 +59,7 @@ def create_app(ignore_exceptions=None, debug=False, **config):
     def capture_exception():
         try:
             raise ValueError('Boom')
-        except:
+        except Exception:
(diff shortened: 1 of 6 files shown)


tests/contrib/flask/tests.py
```diff
@@ -45,7 +45,7 @@ def an_error():
     def capture_exception():
         try:
             raise ValueError('Boom')
-        except:
+        except Exception:
```
mattrobenolt: Linter yelled about this.

dcramer:
@nateberkopec do you remember what we did in raven-ruby for this?

dcramer:
Also if we do this we should update client-dev docs to suggest this

fletom:
thanks @mattrobenolt!

this will make our error data much more useful - it'll align our frontend (raven.js) user IPs with our API (raven-python) user IPs even though we're behind both CloudFront and Heroku

fletom:
@mattrobenolt @dcramer any updates on this? it's a very important feature for us

nateberkopec:
@dcramer here you go: https://github.com/getsentry/raven-ruby/blob/master/lib/raven/utils/real_ip.rb

if-fi:
HI, @mattrobenolt @dcramer I'd like to ask if you are planning to merge this PR soon?

mitsuhiko:
@ashwoods we should merge this after fixing the tests.

PR #977, opened 2017-03-02
No longer provide form data for Flask. Fixes #457

### LINKED ISSUE #457: Prevents access to request raw data in Flask
This all seems to be a bit of a mess, but if you want to access the raw request data in Flask regardless of the content type (I can't control the client), you need to call

```
 request.get_data()
```

and you need to do this before accessing request.data or request.form, because both of these will call get_data() in such a way that the stream will be exhausted, parsed into the form dict, and the raw data is never cached.

This doesn't seem documented except in the comments: https://github.com/mitsuhiko/werkzeug/blob/master/werkzeug/wrappers.py#L440

Should the sentry extension call request.get_data() first to make sure the raw data is cached before using request.form?

mitsuhiko:
I think this feature does not make much sense, particularly because we badly
trim down the form data on the server anyways for security and size reasons.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,8 @@
+Version 6.1.0
+-------------
+
+* No longer provide form data for Flask (#457)
+
 Version 6.0.0
 -------------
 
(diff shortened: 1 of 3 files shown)


mitsuhiko:
@dcramer there are some ways to do it but it is not a stable API in Flask. Ideally we could access the cache value in case it has been pulled in before.

mitsuhiko:
Makes we wonder what we do in Django. That should in theory have the same issue.

PR #978, opened 2017-03-09
Add sample_rate configuration

dcramer:


docs/advanced.rst
```diff
@@ -175,6 +175,16 @@ The following are valid arguments which may be passed to the Raven client:
...
 
+.. describe:: sample_rate
+
+    The sampling factor to apply to events. A value of 0.00 will deny sending
+    any events, and a value of 1.00 will send 100% of events.
+
+    .. code-block:: python
+
+        # send 50% of events
+        sample_rate = 5.0
```
mattrobenolt: Doesn't this contradict what the description says? 0.5 would be sending 50% of events. I didn't read the code to see how it' simplemented though, so not sure which one is accurate.

raven/base.py
```diff
@@ -627,7 +631,9 @@ def capture(self, event_type, data=None, date=None, time_spent=None,
             event_type, data, date, time_spent, extra, stack, tags=tags,
             **kwargs)
 
-        self.send(**data)
+        # should this event be sampled?
+        if self._random.random() <= self.sample_rate:
```
tkaemming: Technically [this is lower bound inclusive](https://docs.python.org/2/library/random.html#random.random), so a sample rate of `0.0` can still result in an event being sent. It'd be safer/more correct to make this comparison `<`. :nerd_face:

Issue #323, opened 2013-06-14
Add sampling


dcramer:
Add an option to set a sample rate for errors

xordoquy:
Is it still something you're willing to do ?

PR #987, opened 2017-04-03
Add provision to sample per message

arnavk:
- Add sample_rate to Client.capture that overrides Client.sample_rate
- Add logic in SentryHandler to look for `sample_rate` in `extra` extracted from the `LogRecord` and pass that to `capture`

Usage: 

```python
some_logger.warning('foo', extra={'sample_rate': 0.5}  # sample 50% of the events

# or

client.captureMessage('Something went fundamentally wrong', sample_rate=0.5)
```

Hi! Would this be something you'd consider adding? If so, I can go ahead and add some docs. If not, is there a variation of this that you'd consider? This could be an alternative (or rather supplementary to client-level sample rate approach) to tackle https://github.com/getsentry/raven-python/issues/323)

### DIFF
--- a/raven/base.py
+++ b/raven/base.py
@@ -566,7 +566,8 @@ class Client(object):
         })
 
     def capture(self, event_type, data=None, date=None, time_spent=None,
-                extra=None, stack=None, tags=None, **kwargs):
+                extra=None, stack=None, tags=None, sample_rate=None,
+                **kwargs):
         """
         Captures and processes an event and pipes it off to SentryClient.send.
 
@@ -636,11 +637,8 @@ class Client(object):
             **kwargs)
 
         # should this event be sampled?
-        sample_rate = self.sample_rate
-        try:
-            sample_rate = float(extra['sample_rate'])
-        except (TypeError, KeyError, ValueError):
-            pass
+        if sample_rate is None:
+            sample_rate = self.sample_rate
 
         if self._random.random() < sample_rate:
             self.send(**data)
--- a/raven/base.py
+++ b/raven/base.py
@@ -636,7 +636,13 @@ class Client(object):
             **kwargs)
 
         # should this event be sampled?
-        if self._random.random() < self.sample_rate:
+        sample_rate = self.sample_rate
+        try:
+            sample_rate = float(extra['sample_rate'])
+        except (TypeError, KeyError, ValueError):
+            pass
+
+        if self._random.random() < sample_rate:
             self.send(**data)
 
         self._local_state.last_event_id = data['event_id']
--- a/raven/handlers/logging.py
+++ b/raven/handlers/logging.py
@@ -174,7 +174,14 @@ class SentryHandler(logging.Handler, object):
         tags.update(getattr(record, 'tags', {}))
 
         kwargs.update(handler_kwargs)
+        sample_rate = extra.pop('sample_rate', None)
+        try:
+            if sample_rate is not None:
+                sample_rate = float(sample_rate)
+        except ValueError:
+            sample_rate = None
 
         return self.client.capture(
             event_type, stack=stack, data=data,
-            extra=extra, date=date, **kwargs)
+            extra=extra, date=date, sample_rate=sample_rate,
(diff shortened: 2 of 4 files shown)


raven/base.py
```diff
@@ -636,7 +636,13 @@ def capture(self, event_type, data=None, date=None, time_spent=None,
             **kwargs)
 
         # should this event be sampled?
-        if self._random.random() < self.sample_rate:
+        sample_rate = self.sample_rate
+        try:
+            sample_rate = float(extra['sample_rate'])
```
dcramer: I would probably do a get here instead of exception and we'd want this to be an arg to capture rather than part of the extra metadata.

arnavk: > I would probably do a get here instead of exception

Do you mean something like: 
`sample_rate = float((extra or {}).get('sample_rate', self.sample_rate))`?

> we'd want this to be an arg to capture rather than part of the extra metadata

hmm, I was thinking the usage would be something like: 
`some_logger.error("asd", extra={'sample_rate': 0.4})`

I'll need a bit of time to dig into the codebase to figure out how to make that work without putting the sample rate in the `extra`

dcramer: if you did that this code would need to live in the logging abstraction, and not here.

the 'extra' bit is 'extra context to send with the event', not to be confused with 'configuration for this event'

arnavk:
@dcramer does this approach make more sense? (Haven't updated the tests yet)

raven/handlers/logging.py
```diff
@@ -174,7 +174,14 @@ def _emit(self, record, **kwargs):
         tags.update(getattr(record, 'tags', {}))
 
         kwargs.update(handler_kwargs)
+        sample_rate = extra.pop('sample_rate', None)
```
arnavk: maybe `get` is a better option? thoughts?

dcramer: i think pop is good because theres no real reason to pass this through contextually if its thought of as configuration

raven/base.py
```diff
@@ -636,7 +637,10 @@ def capture(self, event_type, data=None, date=None, time_spent=None,
             **kwargs)
 
         # should this event be sampled?
-        if self._random.random() < self.sample_rate:
+        if not sample_rate:
```
arnavk: should I add an extra layer of type checking here?

dcramer: oh actually if sample_rate is always None in the logging handler, and thus None gets passed here, that'd be a problem

arnavk: oops... this should have been `if sample_rate is None`

arnavk: if `sample_rate=None` is passed in, it defaults to the client's `sample_rate`... does that sound fair?

dcramer: thats actually probably fine -- if they want to disable sampling they can do sample_rate=1

arnavk: right, so if I just change this to `if sample_rate is None`, that achieves these cases:

sample_rate=None -> default to client.sample_rate
sample_rate=any numeric value -> overwrite client.sample_rate
sample_Rate=1.0 -> disable sampling


Does that sound OK?

dcramer: Yep that sounds good to me.

raven/base.py
```diff
@@ -636,7 +637,10 @@ def capture(self, event_type, data=None, date=None, time_spent=None,
             **kwargs)
 
         # should this event be sampled?
-        if self._random.random() < self.sample_rate:
+        if sample_rate is None:
```
arnavk: thoughts on type checking here? There's no check Client.__init__, so deferring this to the consumer might be fine(?) on the other hand, being defensive isn't the worst idea 🤔

dcramer: i dont think we need type checking -- people can deal with the consequence of putting arbitrary values in

arnavk: ok great. I'll go update the tests then!

dcramer:
Thanks!

arnavk:
Thanks for the guidance, @dcramer! I'll make another PR to add some docs for this.

PR #989, opened 2017-04-04
Add docs to demonstrate sampling messages

arnavk:
- Update Client.capture documentation to describe `sample_rate`
- Add section on sampling in the Advanced Usage docs

This is a follow-up to #987 where we added the provision to sample messages individually using `sample_rate`


dcramer:
Thanks!

Issue #984, opened 2017-03-30
Create a configurable key-sanitizer processor


rz:
I need a processor to sanitize financial data (SSN, EIN, tax ids, bank account numbers, ...). The mechanics of the processor should be very similar to those of [SanitizePasswordsProcessor], except operating on different `FIELDS`.

One potential solution to the general version of this problem is to have a processor that sanitizes a set of configurable keys. The keys to sanitize would be specified with an optional setting.

[SanitizePasswordsProcessor]: https://github.com/getsentry/raven-python/blob/c69abaa1506a426802e979c7f8c6a4ab0ef9c4d7/raven/processors.py#L69

mattrobenolt:
So we could refactor `SanitizePasswordsProcessor` slightly into a generic `SanitizeFieldsProcessor` that takes a set of fields as input. Then `SanitizePasswordsProcesses` ultimately subclasses `SanitizeFieldsProcessor` with a hardcoded list of fields.

Does that make sense?

rz:
Yep, that makes sense. I'll want to have tests for `SanitizeFieldsProcessor` and a code review before refactoring the existing `SanitizePasswordsProcessor`.

PR #990, opened 2017-04-04
Add processor that sanitizes configurable keys

rz:
The logic of the new processor is almost identical to the logic of the
existing SanitizePasswordsProcessor, so this commit also changes the
implementation of SanitizePasswordsProcessor to avoid code-duplication.
It now inherits from the new SanitizeKeysProcessor.

### DIFF
--- a/raven/processors.py
+++ b/raven/processors.py
@@ -68,8 +68,7 @@ class RemoveStackLocalsProcessor(Processor):
 
 class SanitizeKeysProcessor(Processor):
     """
-    Asterisk out things that look like passwords, credit card numbers,
-    and API keys in frames, http, and basic extra data.
+    Asterisk out things that correspond to a configurable set of keys.
     """
     MASK = '*' * 8
 
@@ -144,6 +143,10 @@ class SanitizeKeysProcessor(Processor):
 
 
 class SanitizePasswordsProcessor(SanitizeKeysProcessor):
+    """
+    Asterisk out things that look like passwords, credit card numbers,
+    and API keys in frames, http, and basic extra data.
+    """
     FIELDS = frozenset([
         'password',
         'secret',
--- a/raven/processors.py
+++ b/raven/processors.py
@@ -66,30 +66,23 @@ class RemoveStackLocalsProcessor(Processor):
             frame.pop('vars', None)
 
 
-class SanitizePasswordsProcessor(Processor):
+class SanitizeKeysProcessor(Processor):
     """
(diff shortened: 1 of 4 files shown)


rz:
@mattrobenolt this is my shot at implementing #984.

i think a test to ensure the config is passed-in correctly in all the scenarios may be in order, but i didn't see an obvious place for that. otoh, that change is literally 1 line of code, adding 1 attribute to the Client on init, so maybe it is ok to do without the test.

rz:
@mattrobenolt it would be very awesome if this could make it into an official release soon so that i can install the official version in production for work.

ehfeng:
If we merge this, ping me and I'll add this to our docs.

mattrobenolt:
> If we merge this

and tag a release.

raven/processors.py
```diff
@@ -148,3 +140,31 @@ def _sanitize_keyvals(self, keyvals, delimiter):
...
+        'api_key',
+        'apikey',
+        'sentry_dsn',
+        'access_token',
+    ])
+    VALUES_RE = re.compile(r'^(?:\d[ -]*?){13,16}$')
+
+    def __init__(self, client):
+        super(SanitizeKeysProcessor, self).__init__(client)  # run the __init__ method of Processor, not SanitizeKeysProcessor
+        self.fields = self.FIELDS
```
cancan101: Seems odd to rebind field.

raven/processors.py
```diff
@@ -66,31 +66,23 @@ def filter_stacktrace(self, data, **kwargs):
...
-        'api_key',
-        'apikey',
-        'sentry_dsn',
-        'access_token',
-    ])
-    VALUES_RE = re.compile(r'^(?:\d[ -]*?){13,16}$')
+
+    def __init__(self, client):
+        super(SanitizeKeysProcessor, self).__init__(client)
+        fields = getattr(client, 'sanitize_keys')
```
ashwoods: Could you move this into a `get_fields` method, or rather `get_keys` to be more consistent with naming.

raven/processors.py
```diff
@@ -102,7 +94,7 @@ def sanitize(self, key, value):
             key = text_type(key)
 
         key = key.lower()
-        for field in self.FIELDS:
+        for field in self.fields:
```
ashwoods: this would change to `for field in self.get_keys()`

raven/processors.py
ashwoods: and then it wouldn't be necessary to rebind. just override `get_keys`.

PR #991, opened 2017-04-05
Specify in which version of 2.7 was extra keyword added to .exception

mlazowik:
See tags in the commit: https://github.com/python/cpython/commit/947f358a069c9141c0230e440ab5cba07df0f87f


mattrobenolt:
Sounds legit.

mattrobenolt:
Thanks! 🍦

PR #992, opened 2017-04-06
Add provision to setup Flask logging with filters for SentryHandler

arnavk:
Hi there! Would this be something you're willing to accept?

Add a provision to provide filters for SentryHandler while setting up integration with a Flask app. This is currently possible by poking around the SentryHandler after it has been configured in the `init_app` step but this patch could (maybe) make things a little convenient. If yes, then I can add tests and docs as well.

Usage:
```python
my_filter = SomeLoggingFilter()
sentry.init_app(app, logging_filters=my_filter)
```

### DIFF
--- a/raven/contrib/flask.py
+++ b/raven/contrib/flask.py
@@ -94,13 +94,14 @@ class Sentry(object):
     # gets shared by every app that does init on it
     def __init__(self, app=None, client=None, client_cls=Client, dsn=None,
                  logging=False, logging_exclusions=None, level=logging.NOTSET,
-                 wrap_wsgi=None, register_signal=True):
+                 wrap_wsgi=None, register_signal=True, logging_filters=None):
         if client and not isinstance(client, Client):
             raise TypeError('client should be an instance of Client')
 
         self.dsn = dsn
         self.logging = logging
         self.logging_exclusions = logging_exclusions
+        self.logging_filters = logging_filters
         self.client_cls = client_cls
         self.client = client
         self.level = level
@@ -236,7 +237,7 @@ class Sentry(object):
 
     def init_app(self, app, dsn=None, logging=None, level=None,
                  logging_exclusions=None, wrap_wsgi=None,
-                 register_signal=None):
+                 register_signal=None, logging_filters=None):
         if dsn is not None:
             self.dsn = dsn
 
@@ -262,6 +263,12 @@ class Sentry(object):
         if logging_exclusions is not None:
             self.logging_exclusions = logging_exclusions
 
+        if logging_filters is not None:
+            if not isinstance(logging_filters, list):
+                logging_filters = [logging_filters]
+
+            self.logging_filters = logging_filters
+
         if not self.client:
             self.client = make_client(self.client_cls, app, self.dsn)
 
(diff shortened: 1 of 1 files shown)


arnavk:
any opinions?

dcramer:
I think it'd be better if we sent a signal, or something similar, so that you could do whatever you wanted here vs a one-off for filters.

raven/contrib/flask.py
```diff
@@ -270,7 +277,11 @@ def init_app(self, app, dsn=None, logging=None, level=None,
             if self.logging_exclusions is not None:
                 kwargs['exclude'] = self.logging_exclusions
 
-            setup_logging(SentryHandler(self.client, level=self.level), **kwargs)
+            handler = SentryHandler(self.client, level=self.level)
+            setup_logging(handler, **kwargs)
+            if logging_filters is not None:
+                for _filter in logging_filters:
```
mitsuhiko: why is this called `_filter` and not `filter`?

arnavk: Oops, should have been `filter`

arnavk:
@dcramer did you mean send a signal that the SentryHandler was added (with a reference to the handler?)

dcramer:
@arnavk it could just be a signal that setup_logging was called with the details. Then you can do whatever you want. Its a lot more flexible and still very easy to instrument.

raven/contrib/flask.py
arnavk: I remember now...`filter` is a function... so I just avoided that. Are you ok with `filter` as a name here or should I use `f` or something?

arnavk:
@dcramer If I understand you correctly, once this signal is written, this PR should be replaced by a signal handler like so:

```
def init_app_called_handler(self, app, logging=None, **kwargs):
    if logging is not None:
        self.logging = logging

    if logging_exclusions is not None:
        self.logging_exclusions = logging_exclusions

    if logging_filters is not None:
        if not isinstance(logging_filters, list):
            logging_filters = [logging_filters]

        self.logging_filters = logging_filters
    if self.logging:
        kwargs = {}
        if self.logging_exclusions is not None:
            kwargs['exclude'] = self.logging_exclusions

        handler = SentryHandler(self.client, level=self.level)
        setup_logging(handler, **kwargs)
        if logging_filters is not None:
            for _filter in logging_filters:
                handler.addFilter(_filter)

```

Is that accurate?

dcramer:
@arnavk I'm suggesting firing a signal so anyone can deal with it in whatever way they want, since this filters thing is fairly specific and we get to a point where configuration is too complex to solve ever need.

There's a couple of approaches we could do, though some have complexity:

1. We could fire a signal when Raven is configured
2. We could fire a signal when setup_logging is called
3. Flask-only: We could fire a signal in one or both of the above

The problem with 1/2 is that we'd need our own signal framework in place. It's not a complex thing, but it's probably overkill at this point. Whereas with #3, we can just require blinker like we already do, and wire, at the very least, setup_logging to fire something like a logging_configured signal. Your app could then just bind to that and add any filters/whatever from there.

arnavk:
Ok, I'll rework this with #3. Since it'll likely be a bigger change, I'll make a separate PR and I'm closing this one for now. Thanks, @dcramer!

PR #1007, opened 2017-05-24
Ensure consistent fingerprint for SoftTimeLimitExceeded exceptions

adamcharnock:
I've been finding that ``SoftTimeLimitExceeded`` exceptions have not been grouped in the Sentry UI as I would have expected. I believe I have traced it down to the line changed in this pull request. They seemed to occasionally stop form new groups for non-obvious reasons.

After looking around, it seems that ``sender`` (presumably a task instance) includes an object ID in its string representation. As a result, the fingerprint would be different for every celery worker process.

This change will attempt to use the full task name (i.e. ``"my.app.tasks.do_it"``) before falling back to the original implementation.

### DIFF
--- a/raven/contrib/celery/__init__.py
+++ b/raven/contrib/celery/__init__.py
@@ -73,7 +73,7 @@ class SentryCeleryHandler(object):
 
         # This signal is fired inside the stack so let raven do its magic
         if isinstance(einfo.exception, SoftTimeLimitExceeded):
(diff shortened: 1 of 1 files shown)


dcramer:
What version of Celery was this on btw?

adamcharnock:
Thanks @dcramer, this was on celery 4.0.2, raven 5.27.0.

PR #1011, opened 2017-05-31
Adding framework name to sdk name

ehfeng:
Doing this for Flask to see if this is the right pattern. I'll fill this in for the rest of the frameworks. This is tied with the FeatureAdoption https://github.com/getsentry/sentry/pull/5481, to track what frameworks are being used.

### DIFF
--- a/raven/base.py
+++ b/raven/base.py
@@ -152,7 +152,7 @@ class Client(object):
     def __init__(self, dsn=None, raise_send_errors=False, transport=None,
                  install_sys_hook=True, install_logging_hook=True,
                  hook_libraries=None, enable_breadcrumbs=True,
-                 _random_seed=None, sdk_name=None, **options):
+                 _random_seed=None, integration_name=None, **options):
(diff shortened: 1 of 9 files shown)


bretthoerner:
Sentry side: https://github.com/getsentry/sentry/pull/5517

ehfeng:
@bretthoerner is it cool if i just complete this PR to add the new "integration" field? just checking in case there's anything I've forgotten. Also, @MaxBittker was asking if there's a canonical list of all platforms/integrations that I should be grabbing strings from.

bretthoerner:
Yeah, `integrations` in the `sdk` object is now valid and takes an array of strings.

There's no canonical list, it's just a "what makes sense for the language" kind of thing. Basically anything that enables special code paths in the SDK, so framework and logging integrations, I'm not sure what else?

ehfeng:
We're changing how this works, using a `platform_id` field instead.

PR #1012, opened 2017-06-07
Get hostname when building message

luqasz:
This patch will get current hostname when message is build
(e.g. on time of sending it). Previous solution used cached hostname
which resulted on wrong hostname being sent while it was changed during
application run time.

### DIFF
--- a/raven/base.py
+++ b/raven/base.py
@@ -15,6 +15,7 @@ import sys
 import time
 import uuid
 import warnings
+import socket
 
 from datetime import datetime
 from inspect import isclass
@@ -445,7 +446,10 @@ class Client(object):
             data['level'] = kwargs.get('level') or logging.ERROR
 
         if not data.get('server_name'):
-            data['server_name'] = self.name
+            if hasattr(socket, 'gethostname'):
+                data['server_name'] = socket.gethostname()
+            else:
+                self.name
 
         if not data.get('modules'):
             data['modules'] = self.get_module_versions()


mattrobenolt:
In what scenario does the hostname change during runtime?

luqasz:
> In what scenario does the hostname change during runtime?


On instance boot from custom AMI image on AWS. How to reproduce (without AWS):

1. launch ipython (or whatever you like)
2. setup raven client with valid DSN
3. run 
```python
try:
    1 / 0
except ZeroDivisionError:
    client.captureException()
```
server name is exactly the same as hostname on which ipython is executed
4. change hostname by (with root privileges) `hostname some_name` 
5. run
```python
try:
    1 / 's'
except TypeError:
    client.captureException()
```
exception sent to sentry does still has old server name value

mattrobenolt:
Yeah, I get _how_ it can change, I just am trying to understand how while a process is already up and running, it's hostname would change from under it. Checking hostname requires an additional syscall. So introducing this is kinda silly for everyone for a situation that I find hard to believe is common, considering this is the first I've heard of it. I'd also argue that changing the hostname from under a running process would likely cause other issues on the server.

luqasz:
> I'd also argue that changing the hostname from under a running process would likely cause other issues on the server.

What do you mean by that ? Change hostname from application itself ?

> Checking hostname requires an additional syscall. So introducing this is kinda silly for everyone for a situation that I find hard to believe is common

Yes it is a uncommon situation. Still possible.

mattrobenolt:
> What do you mean by that ? Change hostname from application itself ?

I mean, something is changing the hostname while the process is running. I don't know if it's the process itself, or some outside force. My argument is that anything doing this is probably not a great idea and will potentially cause other issues.

So you're saying that when you spawn a new instance in AWS, you have one hostname, I'm guessing was baked into the AMI, then during boot, your application starts up, _then_ `cloud-init` kicks in later, changing the hostname?

luqasz:
> So you're saying that when you spawn a new instance in AWS, you have one hostname, I'm guessing was baked into the AMI, then during boot, your application starts up, then cloud-init kicks in later, changing the hostname?

Yes. We will reorder init scripts as we can so that hostname will not change after application boot up. 

Maybe a user defined callable for `Client` class `__init__()` will do. It would default to `raven.conf.defaults.NAME` but when a user passes a callable then it will be executed and its return value used as a server name.

ashwoods:
Think I'm going to close this for now. If we still have issues that might warrant a change we will search for a solution.

luqasz:
>Checking hostname requires an additional syscall. So introducing this is kinda silly for everyone for a situation that I find hard to believe is common

Using http/https transport, raven will open a new connection either way which is way more costly than simple `socket.gethostname()`.

PR #1014, opened 2017-06-07
Added support for logging integration via ZConfig:

jimfulton:
See: http://zconfig.readthedocs.io/en/latest/using-logging.html

Note that the Zope integration also uses ZConfig, but adds lots of Zope-specific machinery that won't work elsewhere.  This integration cribbed from the Zope integration. :)

### DIFF
--- a/docs/integrations/zconfig.rst
+++ b/docs/integrations/zconfig.rst
@@ -3,8 +3,9 @@ ZConfig logging configuration
 
 `ZConfig
 <http://zconfig.readthedocs.io/en/latest/using-logging.html>`_
-provides one of the most powerful clean configuration mechanism for
-the Python logging module.  To learn more, see:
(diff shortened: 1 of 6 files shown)


jimfulton:
I've cleaned up the implementation a little and added tests.

docs/integrations/zconfig.rst
```diff
@@ -0,0 +1,32 @@
+ZConfig logging configuration
+=============================
+
+`ZConfig
+<http://zconfig.readthedocs.io/en/latest/using-logging.html>`_
+provides one of the most powerful clean configuration mechanism for
```
dcramer: it probably makes sense to keep the description unbiased:

> ZConfig provides a configuration mechanism for the Python logging module

jimfulton:
OK. language changed.

dcramer:
thanks!

PR #1017, opened 2017-06-13
Fix JSON encoding of non-standard dict keys

### LINKED ISSUE #982: Sentry fall if there is a frozenset as key of dict in context
Sentry fall if there is a ```frozenset``` as key of ```dict``` in context
I found this  when near place of send error was ```dict``` with ```frozenset``` as key

proof 1
```
from  raven.utils.json import *
value = {frozenset([1,2,3]):4}
json.dumps(value, cls=BetterJSONEncoder)
  File "<input>", line 1, in <module>
  File "/usr/lib/python2.7/json/__init__.py", line 251, in dumps
    sort_keys=sort_keys, **kw).encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
TypeError: keys must be a string
```

user case
```
/pycharm-2016.3.2/helpers/pydev/pydevconsole.py 39277 38925
PyDev console: starting.
import sys; print('Python %s on %s' % (sys.version, sys.platform))
import django; print('Django %s' % django.get_version())
sys.path.extend(['/home/uc/Projects/c2public', '/home/uc/Downloads/pycharm-2016.3.2/helpers/pycharm', '/home/uc/Downloads/pycharm-2016.3.2/helpers/pydev'])
if 'setup' in dir(django): django.setup()
import django_manage_shell; django_manage_shell.run("/home/uc/Projects/c2public")
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
Django 1.9.9
log = logging.getLogger('django')
  File "<input>", line 1, in <module>
NameError: name 'logging' is not defined
import logging
log = logging.getLogger('django')
log.error("test error 1")
ERROR test error 1
a = {frozenset([1,2,3]):4}
log.error("test error 2")
ERROR test error 2
log.error("test error 3", extra={'locals': locals()})
ERROR test error 3
Top level Sentry exception caught - failed creating log record
test error 3
  File "/home/uc/envs/c2public/local/lib/python2.7/site-packages/raven/handlers/logging.py", line 89, in emit
    return self._emit(record)
  File "/home/uc/envs/c2public/local/lib/python2.7/site-packages/raven/contrib/django/handlers.py", line 33, in _emit
    return super(SentryHandler, self)._emit(record, request=request)
  File "/home/uc/envs/c2public/local/lib/python2.7/site-packages/raven/handlers/logging.py", line 180, in _emit
    extra=extra, date=date, **kwargs)
  File "/home/uc/envs/c2public/local/lib/python2.7/site-packages/raven/contrib/django/client.py", line 284, in capture
    result = super(DjangoClient, self).capture(event_type, **kwargs)
  File "/home/uc/envs/c2public/local/lib/python2.7/site-packages/raven/base.py", line 610, in capture
    self.send(**data)
  File "/home/uc/envs/c2public/local/lib/python2.7/site-packages/raven/base.py", line 703, in send
    message = self.encode(data)
  File "/home/uc/envs/c2public/local/lib/python2.7/site-packages/raven/base.py", line 745, in encode
    return zlib.compress(json.dumps(data).encode('utf8'))
  File "/home/uc/envs/c2public/local/lib/python2.7/site-packages/raven/utils/json.py", line 51, in dumps
    return json.dumps(value, cls=BetterJSONEncoder, **kwargs)
  File "/usr/lib/python2.7/json/__init__.py", line 251, in dumps
    sort_keys=sort_keys, **kw).encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
TypeError: keys must be a string


```

smokey42:
* Fixes #982 (any my own problem)
 * Enables `defaultdict` instances to be passed
 * reason this occurred was that the crash occurred
   before `default` was actually called, so the
   `BetterJSONEncoder` was not able to catch this.
 * Added plenty of tests for those and similar cases.

### DIFF
--- a/raven/utils/json.py
+++ b/raven/utils/json.py
@@ -9,6 +9,7 @@ raven.utils.json
 from __future__ import absolute_import
 
 import codecs
+import collections
 import datetime
 import uuid
 import json
@@ -27,6 +28,30 @@ class BetterJSONEncoder(json.JSONEncoder):
         frozenset: list,
         bytes: lambda o: o.decode('utf-8', errors='replace')
     }
+
+    def encode(self, obj):
+        super_encode = super(BetterJSONEncoder, self).encode
+        try:
+            return super_encode(obj)
+        except TypeError:
+            # json.encode keeps crashing somewhere in the C code called by
+            # `iterencode` before `default` can actually be called.
+            # We need to massage the data a bit and try again.
+            return super_encode(self.encode_keys(obj))
+
+    def encode_keys(self, value):
+        # Need to do this recursively, though this is the last resort anyways.
+        # Alternative would be to crash, but this is not really an alternative.
+        if isinstance(value, collections.Mapping):
+            return {self.encode_keys(key): val
+                    for key, val in value.items()}
+        elif isinstance(value, frozenset):
+            return repr(value)
+        else:
+            try:
+                return self.default(value)
+            except TypeError:
+                return repr(value)
 
     def default(self, obj):
         try:
--- a/tests/utils/json/tests.py
+++ b/tests/utils/json/tests.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+import collections
 import datetime
 import uuid
 from decimal import Decimal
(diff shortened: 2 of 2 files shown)


dcramer:
How is a defaultdict getting all the way to the JSON serializer? These should get caught by the other serializer code I thought.

smokey42:
@dcramer By passing it e.g. via `logger.warn("Foo", extra=locals())`. I remembered I had an issue with `defaultdict` before, can't remember if it was passed directly or via `locals()` though.

dcramer:
I'm not opposed to this, but if thats the case, it seems like we'd be better off trying to correct whatever broke within logger (could have been something like tags or similar). There's an expectation that things passed to our JSON serializer are already JSON-compatible types.

smokey42:
Does this mean that it's the developers problem if she uses `logger.error("Foo", extra=locals())` and having a `dict` with a non-serializable key somewhere in there or that there is something wrong with other parts of Raven having it propagate this far in the first place? If so can you please point me to the relevant part of the code so I can have a look at that?

dcramer:
@smokey42 I think my preferred approach would be that in our logging handler we ensure that data passed from extra to whatever is not an invalid type. I'm imagining this is a specific attribute that isnt coerced correctly (e.g. tags or something).

smokey42:
@dcramer So if you imagine never passing invalid stuff to the JSON encoder, would you like the `BetterJSONEncoder` being removed as well and have some coercion function used instead on all call sites? A quick `grep` only shows `Client.encode`. As the coercion should be done as late as possible (but not in `BetterJSONEncoder` which is the latest possible) it imho should be done in `Client.encode`.

PR #1022, opened 2017-06-23
Purge scheme from dsn templatetag

beck:
As is, when capturing an exception on http, the proto relative call results in:

> 307 Internal Redirect
> Location: https://app.getsentry.com/api/....

Which is troublesome because it results in a CORS error being logged:

> XMLHttpRequest cannot load .... Origin 'null' is therefore not allowed access.

With sentry enforcing https, the templatetag is a bit broken. It should *always* use https.

The event is is still successfully logged.

The error is, at best, annoying. Or if you're a dope like me (and mired in IE debuggin) you get side-tracked by such an error and start pointing fingers at sentry for things not working (which is just a red herring).

This change enforces https and simplifies matters by purging the `scheme` argument.

Sidenote about the doc update:

A strict Content Security Policy will disallow inline scripting, the updated example allows for the templatetag to be consumed using an external javascript file.

### DIFF
--- a/docs/integrations/django.rst
+++ b/docs/integrations/django.rst
@@ -57,15 +57,13 @@ Inside your template, you can now use:
 
 .. sourcecode:: html+django
 
-    <script>Raven.config('{% sentry_public_dsn %}').install()</script>
+    <body data-sentry="{% sentry_public_dsn %}">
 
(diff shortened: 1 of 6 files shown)


beck:
Gah, django blows up if you continue to pass a scheme.

Will return when this doesn't break backward compatibility so terribly.

PR #1023, opened 2017-06-23
Default `get_public_dsn()` to https 

beck:
The API enforces https, the DSN returned should match.

The issue with proto-relative is that it creates CORS errors.

On http, the call results in:

> 307 Internal Redirect
> Location:https://app.getsentry.com/api/....

and the console errors:

> XMLHttpRequest cannot load .... Origin 'null' is therefore not allowed access.

The event is is still successfully logged.

The error is, at best, annoying. At worst, if you're a dope like me (stumped by some IE bug) you let the error become a red herring and blame raven-js for your woes.

The `scheme` argument can not be purged without breaking backward compatibility. Folks already using:

```
{% sentry_public_dsn 'https' %}
```

will see no change. Those using:

```
{% sentry_public_dsn %}
```

Will no longer get a console error.

### DIFF
--- a/docs/integrations/django.rst
+++ b/docs/integrations/django.rst
@@ -57,15 +57,13 @@ Inside your template, you can now use:
 
 .. sourcecode:: html+django
 
-    <script>Raven.config('{% sentry_public_dsn %}').install()</script>
+    <body data-sentry="{% sentry_public_dsn %}">
 
-By default, the DSN is generated in a protocol relative fashion, e.g.
-``//public@example.com/1``. If you need a specific protocol, you can
-override:
+Inside your javascript:
 
-.. sourcecode:: html+django
+.. sourcecode:: js
 
-    {% sentry_public_dsn 'https' %}
+    Raven.config(document.body.dataset.sentry).install();
 
 .. sentry:edition:: hosted, on-premise
 
--- a/docs/integrations/flask.rst
+++ b/docs/integrations/flask.rst
@@ -161,7 +161,7 @@ crash dialog:
     def internal_server_error(error):
         return render_template('500.html',
             event_id=g.sentry_event_id,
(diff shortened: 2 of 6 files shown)


mattrobenolt:
So to comment here on some high level things without actually reviewing code:

* The redirect you’re seeing is coming from your browser, not our server. We do allow ingestion for this endpoint alone over both http and https. The protocol needs to match to avoid cross-origin requests, which strip headers that we need.
* Regardless of this change, raven-js takes it upon itself to strip the protocol anyways, no matter what we specify here for the above issue. So in theory, no change here will actually have an impact.
* Users run Sentry themselves in an environment that might not even have https, so we can enforce this on them.

beck:
> The redirect you’re seeing is coming from your browser, not our server.

Is this entirely accurate?

When I do:
```
curl -I https://app.getsentry.com
Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
```

So yeah, the browser is redirecting, but only because the server is setting HSTS.

> We do allow ingestion for this endpoint alone over both http and https

How is http possible with HSTS? I'm assuming outside of a browser context? ... now curious if the template tag is ever consumed for a non-browser environment.

> Users run Sentry themselves in an environment that might not even have https, so we can enforce this on them.

👍 

Thanks for the reply.

beck:
> Regardless of this change, raven-js takes it upon itself to strip the protocol anyways

Double check this? I'm currently using raven-js@3.7

Updated to production to use `{% sentry_public_dsn 'https' %}`, I see `Raven._dsn` is https and all CORS now happen without error.

With the service using HSTS (and the web trending toward https) I think there is still value in this pull. Granted, http users would now have to pass the scheme `http` (which may be an intolerable break in compatibility).

PR #1028, opened 2017-06-26
Fire signal after setup_logging is called for Flask usage

arnavk:
Fire a blinker signal after `setup_logging` is called. 

Replaces #992 

cc @dcramer

### DIFF
--- a/raven/contrib/flask.py
+++ b/raven/contrib/flask.py
@@ -17,6 +17,7 @@ else:
 
 import logging
 
+import blinker
 from flask import request, current_app, g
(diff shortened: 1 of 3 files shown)


arnavk:
Questions:
- Should I move the signal definition to within `raven/contrib/flask.py`?
- Should blinker now be a strong requirement?

dcramer:
@arnavk i think it'd be nice to have signals in general, but given we dont bundle dependencies with raven I'd say just move it to Flask-only for now

arnavk:
Done. Does this look ok? If so, I can add some documentation about this before we merge this

Issue #1032, opened 2017-06-30
Import Behavior / Recursive Errors


dcramer:
We're specifically seeing this in Sentry, but it means its possible elsewhere to. The issue is with importing ``django.conf.settings`` in the same module that has the implementation of ``SENTRY_CLIENT`` defined.

```
:56:45 server          | Python auto-reloader enabled
16:56:45 server          |   File "/Users/dcramer/Development/sentry/src/sentry/wsgi.py", line 22, in <module>
16:56:45 server          |     configure()
16:56:45 server          |   File "/Users/dcramer/Development/sentry/src/sentry/runner/__init__.py", line 123, in configure
16:56:45 server          |     configure(ctx, py, yaml, skip_service_validation)
16:56:45 server          |   File "/Users/dcramer/Development/sentry/src/sentry/runner/settings.py", line 153, in configure
16:56:45 server          |     }, skip_service_validation=skip_service_validation)
16:56:45 server          |   File "/Users/dcramer/Development/sentry/src/sentry/runner/initializer.py", line 243, in initialize_app
16:56:45 server          |     configure_structlog()
16:56:45 server          |   File "/Users/dcramer/Development/sentry/src/sentry/runner/initializer.py", line 235, in configure_structlog
16:56:45 server          |     logging.config.dictConfig(settings.LOGGING)
16:56:45 server          |   File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/config.py", line 794, in dictConfig
16:56:45 server          |     dictConfigClass(config).configure()
16:56:45 server          |   File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/config.py", line 559, in configure
16:56:45 server          |     'filter %r: %s' % (name, e))
16:56:45 server          | ValueError: Unable to configure filter 'sentry:internal': 'module' object has no attribute 'SentryInternalClient'
16:56:45 server          | unable to load app 0 (mountpoint='') (callable not found or import error)
```

Trying to debug this real quick to see what we changed to cause this, but it was something from the 6.0.0 release AFAIK.

dcramer:
Changed import logic in 6.0:

```diff 
+from django.utils.datastructures import MultiValueDict
-from raven.contrib.django.middleware import SentryLogMiddleware
+from raven.contrib.django.middleware import SentryMiddleware
[...]

PR #1033, opened 2017-06-30
Fix import locking to avoid recursion (fixes GH-1032)

dcramer:


mitsuhiko:
This might actually need an actual lock. But this makes sense.

dcramer:
@mitsuhiko it has a lock as well

mitsuhiko:
@dcramer sorry, I meant a recursive lock. (`RLock` instead of `Lock`).

Issue #1051, opened 2017-08-04
Expect logging._srcfile to be None


MarcSchmitzer:
The `breadcrumbs` module uses the `logging._srcfile` attribute and crashes if it is `None`. The `logging` module sort of allows this, though:

> \# _srcfile is only used in conjunction with sys._getframe().
> \# To provide compatibility with older versions of Python, set _srcfile
> \# to None if _getframe() is not available; this value will prevent
> \# findCaller() from being called.

We use a third party library that actively sets `logging._srcfile` to `None`. I'm aware that asking raven to consider this is rather weird, but I thought I'd at least ask :smile: 

I can provide a simple patch that makes `_wrap_logging_method` bail in that case, if there is interest.

dcramer:
IMO our SDK should always be defensive. I'm not sure if this is causing a hard error but:

- it should only log errors and never fail
- we should support odd environments to the best of our ability (do a patch here is fine)

ashwoods:
Fixed in #1053

PR #1052, opened 2017-08-07
breadcrumbs: Check logging._srcfile in _patch_logger

MarcSchmitzer:
The _srcfile attribute of the logging module may be None if the
interpreter doesn't support frame introspection (see comment in
logging/__init__.py and #1051).

### DIFF
--- a/raven/breadcrumbs.py
+++ b/raven/breadcrumbs.py
@@ -200,6 +200,13 @@ def _wrap_logging_method(meth, level=None):
 
 
 def _patch_logger():
+    if logging._srcfile is None:
+        logger.warning(
+            'Frame introspection not available. '
+            'Cannot instrument logging for breadcrumbs.'
+        )
+        return
+
     cls = logging.Logger
 
     methods = {


MarcSchmitzer:
I tried adding a test for this, but gave up because there is so much global state involved.

The check is in `_patch_logger` (not `_wrap_logging_method`) to avoid repeating the warning message for every method being patched.

ashwoods:
@MarcSchmitzer would using `os.path.normcase(logging.currentframe.__code__.co_filename)` instead of using `logging._srcfile` work in your case?

MarcSchmitzer:
@ashwoods That does yield a valid filename and avoids the crash, yes. I'm not sure how to tell whether the `_wrap_logging_method` still has the intended effect with that, though.

ashwoods:
considering that logging._srcfile [defaults to currentframe.__code__.co_filename](https://github.com/python/cpython/blob/75f0b5dbac3376a3b36c943ec867c0daed35eb4f/Lib/logging/__init__.py#L178) , it should allow the breadcrumb code to work. Afaik we don't call getframe in the code, for which logging._srcfile is a sort of switch, just using srcfile to trick the logging module where the patch originates from.  I think we can safely assume to use that as a default if logging._srcfile is None.

MarcSchmitzer:
Ok. I've cooked up a new branch that uses a fallback for `logging._srcfile': https://github.com/solute/raven-python/tree/breacrumbs-logging-srcfile-none2

Should I make a new PR for that?

ashwoods:
@MarcSchmitzer sure.

MarcSchmitzer:
Replaced by #1053 .

Issue #1064, opened 2017-08-22
[flask] Resource leak when register_signal=False


bijanvakili:
For the `raven.contrib.flask.Sentry` client, the documentation explains the `register_signal` parameter as follows:
```
Hook into the `got_request_exception` signal. This can be disabled by passing `register_signal=False`.
```

However, when setting `register_signal=False`, the `init_app` method [also disables registering the 'after_request' handler](https://github.com/getsentry/raven-python/blob/master/raven/contrib/flask.py#L291) despite the documentation not indicating so.  This causes a major resource leak if the `handle_exception()` method is called explicitly since the client `context` is never cleared.

The `after_request` handler should be hooked into the Flask [request_finished](http://flask.pocoo.org/docs/0.12/api/#flask.request_finished) signal in all cases.

PR #1065, opened 2017-08-22
Fix issue #1064: Resource leak when register_signal=False

bijanvakili:
Memory leak fix for issue #1064 

Fix the `raven.contrib.flask.Sentry` client so that the `after_request` handler is called to release all resources in all cases.


dcramer:
i wonder if the reason we did this was blinker becomes optional if you do register_signal=False?

either way lgtm

PR #1072, opened 2017-09-07
Added Django setting to ignore expected Celery errors

dhoffman34:
After installing Raven by adding `raven.contrib.django.raven_compat` to `INSTALLED_APPS` in Django, `SentryCeleryHandler` gets set up, but there is no option to ignore expected exceptions for Celery, even though that is something that `SentryCeleryHandler` has an option for.

I have updated this to add that to be configured as a Django setting. The reason is that we were running into an issue where many expected errors were being reported and we couldn't find a way to stop it apart from monkeypatching `SentryDjangoHandler`.

### DIFF
--- a/docs/integrations/django.rst
+++ b/docs/integrations/django.rst
@@ -296,6 +296,12 @@ Additional Settings
             'CELERY_LOGLEVEL': logging.INFO
         }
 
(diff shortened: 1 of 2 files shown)


dcramer:
Thanks!

PR #1077, opened 2017-09-12
Travis deployments

ashwoods:
Deployments to pypi on pushed git tags


ashwoods:
hmm, maybe I should test first with another script. I might have to add another stage with a condition.

ashwoods:
No need for conditions on its own stage.

ashwoods:
Damn, travis is always tricky. @dcramer the current config with build stages works ok.

mitsuhiko:
@ashwoods I'm good with this however I would propose to pass `--format=zip` to bdist_wheel to make windows users a bit more happy.

ashwoods:
@mitsuhiko I guess you meant to sdist right? afaik isn't wheel always a zip?

mitsuhiko:
Sorry yes, i mean sdist.

PR #1119, opened 2017-10-29
stacks: Fix "filename" not being relative for traces generated on Windows

lazka:
Instead of hardcoding "/" use os.sep


lazka:
Thanks!

PR #1121, opened 2017-10-29
ci: Fix deploy stage

jan-auer:


.travis.yml
```diff
@@ -137,8 +137,6 @@ jobs:
         bucket: getsentry-builds
         upload-dir: $TRAVIS_REPO_SLUG/$TRAVIS_COMMIT
         local_dir: build
-        on:
-          tags: true
```
jan-auer: According to the [docs](https://docs.travis-ci.com/user/deployment) the branch doesn't have to be specified if its `master`

PR #1130, opened 2017-11-07
doco for user context and tags for logging integrations

bendog:
Thought I'd help out and add the tags and user context syntax to the logging integrations docs


ashwoods:
Included changes in pr1131

PR #1134, opened 2017-11-12
Add the Client.install_asyncio_hook function.

Kentzo:
asyncio needs special handling as its unhandled exceptions are consumed by logging and ignoring them.

@asvetlov Please take a look at the implementation and let me know if handling can be improved. I used implementation of `BaseEventLoop.default_exception_handler` as a reference but didn't check all available context's keywords in all released versions of Python.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -302,3 +302,19 @@ A Note on uWSGI
 If you're using uWSGI you will need to add ``enable-threads`` to the
 default invocation, or you will need to switch off of the threaded default
 transport.
+
+Integration with asyncio
+---------------
+
+asyncio introduces additional level of exception handling: exceptions emitted by tasks and futures
+must be explicitly handled and are not automatically propagated into ``sys.excepthook``.
+Unhandled exceptions (exceptions that were not retrieved before object's destruction) are passed into
+loop's own exception handler.
+
+To handle them the per-loop exception handler can be installed:
+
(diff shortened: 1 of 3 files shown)


Kentzo:
I'm not sure how to cause asyncio to call exception handler without 'exception' but with handle of future with the _source_traceback attribute set.

Kentzo:
@ashwoods I used the 'exception' level instead of 'fatal' as handled errors don't explicitly cause interpreter to exit by default. Was it a good idea?

Kentzo:
@ashwoods Is there anything I should change / fix?

asvetlov:
The idea is strange for me.
Why is it a part of `raven-python`? 
I thought `raven-aiohttp` is better place for asyncio-related things.

Kentzo:
@asvetlov This change is to handle asyncio exceptions and unrelated to aiohttp transport.

Kentzo:
@ashwoods I would love to address the issues if there are any

Kentzo:
Anyone? I'm willing to fix any issues.

ashwoods:
I haven't had time to look into context in the loop. Will breadcrumbs leak and be cleared?

Kentzo:
@ashwoods What test would show it?

Kentzo:
Re-opened to re-run CI tests.

@ashwoods Any thoughts?

Issue #1127, opened 2017-11-03
Support for Django 2.0 URL Resolver


ikosenn:
Django 2.0 has changed its routing. The resolver throws an ``Attribute Error`` in a bunch of places.

ikosenn:
![screen shot 2017-11-03 at 01 53 37](https://user-images.githubusercontent.com/5145715/32366484-7e05284c-c03b-11e7-8056-07c4dd06a12b.png)

dkaranja:
+1

khakulov:
Django 2.0 is already relased. This bug is now blocking our upgrade to Django 2.0.

wooyek:
When we can expect a new release with Django 2.0 support?

ashwoods:
@wooyek this week.

wooyek:
@ashwoods is there a new ETA?

wyc902:
Is this issue resolved in django 2.0?

ikosenn:
@wyc902 yes it is. version >=6.4.0

ikosenn:
@wyc902 yes it is. version >=6.4.0

Mee1ad:
django 2.2.3 
same error
AttributeError: 'URLResolver' object has no attribute '_regex'

PR #1149, opened 2017-12-03
Work around a crash in Django 2.0 URLResolver handling

jleclanche:
This is a workaround for #1127... it doesn't fix it, but at least it prevents Raven from crashing, and should fall back on behaviour of the URL resolution being ignored.

Django 2.0 has made fairly extensive changes to URL resolution and I don't fully understand how this handling works in Raven so I don't know if I can actually fix it the right way.


ashwoods:
@jleclanche thx! Might need more testing, but a fix is already in master.

PR #1154, opened 2017-12-05
fix error django VERSION remark

twz915:
Applications
New in Django 1.7.
https://docs.djangoproject.com/en/1.8/ref/applications/#module-django.apps


ashwoods:
@twz915 thx! can you rebase with master so that you get the fixes to the build?

PR #1167, opened 2017-12-19
Add support for asyncio compatible transport.

ticosax:
Rely on aiohttp's client to perform the requests.

Note: only python3.6+ is supported as we don't require to pass the loop to
`asyncio.ensure_future`. Only python3.6 handles this correctly by obtaining
the running loop from the context.

https://docs.python.org/3/whatsnew/3.6.html#asyncio

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -126,6 +126,10 @@ jobs:
       python: 3.6
       env: TOXENV=py36-lambda
 
+    - stage: contrib
+      python: 3.6
+      env: TOXENV=py36-aiohttp
+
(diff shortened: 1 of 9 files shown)


dcramer:
How is this different than raven-aiohttp?

dcramer:
Iirc pip throws compile errors because it’s python 3 only. We’re also trying to move away from “everything must live in one sdk”, though we should at least clearly document it (assuming we don’t).

ticosax:
ok thank you

PR #1170, opened 2017-12-20
folder needs to be created first

friedcell:
on clean clone, .git/hooks does not exist so this fails unless directory is created first

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ coverage:
 
 setup-git:
 	git config branch.autosetuprebase always
+	mkdir -p .git/hooks
 	cd .git/hooks && ln -sf ../../hooks/* ./
 
 publish:


dcramer:
I've never actually seen that behavior before (granted I might have done something else after clone that I dont recall). This would likely cause broken behavior in a number of other repos.

friedcell:
Try a clean clone and then `make test`.

What do you mean by "cause broken behavior in a number of other repos"? The -p prevents an error if the directory is already there, and this is only run when the directory is required in the next step.

ashwoods:
@friedcell what version of git are you using? I just tried a clean `git clone` on my machine and it works. Out of curiosity, as `mkdir -p` won't hurt anyone.

friedcell:
Weird, I'm getting hooks now too. My bad, ignore, not sure why it wasn't there before.

PR #1171, opened 2017-12-20
fix flake8 hook

friedcell:
pre-commit hook (installed with `make setup-git`) seems to be using flake8 2.x api while setup.py requires 3.5.0 - not sure what to do with USER_CONFIG, so left it out.


ashwoods:
Thx, fixed (again) in master.

PR #1180, opened 2018-01-11
Terminate silently when using the threaded transport

gukoff:
Logging this information to stdout is unexpected for users and hard to get rid of.

Libraries shouldn't mess with stdout directly, because:
- it interferes with [logging in 12-factor apps](https://12factor.net/logs);
- such a logging is not configurable;
- the Ctrl+C advice is useless 99% of the time.

### DIFF
--- a/raven/transport/threaded.py
+++ b/raven/transport/threaded.py
@@ -21,6 +21,7 @@ from raven.utils.compat import Queue, check_threads
 DEFAULT_TIMEOUT = 10
 
 logger = logging.getLogger('sentry.errors')
+raven_logger = logging.getLogger('raven')
 
 
 class AsyncWorker(object):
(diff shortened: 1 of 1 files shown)


gukoff:
This PR has been open for 3 years now :) 
Any review?

ashwoods:
@gukoff raven-python has been deprecated in favor of the new python SDK: https://github.com/getsentry/sentry-python. Are there any reasons for keeping you from migrating?

gukoff:
> raven-python has been deprecated in favor of the new python SDK

Ah, true. It's not immediately clear from the README. Seems like the new library has this error fixed.

> Are there any reasons for keeping you from migrating?

No worries, I don't work with Sentry on Python anymore. Just revisiting my open PRs.

PR #1194, opened 2018-02-02
fix(breadcrumbs): Ensure maximum lengths on several attributes

dcramer:
- Trim message, category, and level attributes
- Update pytest support to work in other environments

Fixes GH-604

### DIFF
--- a/raven/breadcrumbs.py
+++ b/raven/breadcrumbs.py
@@ -1,17 +1,21 @@
 from __future__ import absolute_import
 
 import os
-import time
 import logging
+
+from time import time
 from types import FunctionType
 
-from raven.utils.compat import iteritems, get_code, text_type, string_types
 from raven.utils import once
+from raven.utils.encoding import to_unicode
+from raven.utils.compat import iteritems, get_code, text_type, string_types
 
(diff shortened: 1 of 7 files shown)


dcramer:
This also fixes py.test support which had been broken for me locally.

flake8 is still broken (i had to commit with -n) but I dont have time to deal with that. We should be careful changing development/test environment, as it seems we might have done it without considering consequences (e.g. I assume we bumped the version of flake8 and it has never worked w/ our hooks since)

dcramer:
@jan-auer @HazAT throwing you two on since we dont have an sdks team (afaik) on github

raven/breadcrumbs.py
```diff
@@ -28,40 +32,59 @@ def event_payload_considered_equal(a, b):
 
 class BreadcrumbBuffer(object):
 
-    def __init__(self, limit=100):
+    def __init__(self, limit=100, message_max_length=1024):
```
ashwoods: I'll set `limit` and `message_max_length` as settings in the new-settings-branch

ashwoods:
@dcramer Thx!

PR #1201, opened 2018-02-17
Bump references to 2012-2017 to 2018 in author docstrings.

antinescience:
Like @dcramer said, [it's not 2010 any more](https://github.com/getsentry/raven-python/commit/13260b0f5b0e14ce1f3081e5163aaa4f1ee5cec1#diff-f7922c65311be5616a24eea9956f3fe4). This is _super_ pedantic, but it was also bothering the heck out of me. 😁

### DIFF
--- a/raven/contrib/__init__.py
+++ b/raven/contrib/__init__.py
@@ -2,7 +2,7 @@
 raven.contrib
 ~~~~~~~~~~~~~
 
-:copyright: (c) 2010-2013 by the Sentry Team, see AUTHORS for more details.
+:copyright: (c) 2010-2018 by the Sentry Team, see AUTHORS for more details.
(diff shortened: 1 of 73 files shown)


raven/contrib/__init__.py
```diff
@@ -2,7 +2,7 @@
 raven.contrib
 ~~~~~~~~~~~~~
 
-:copyright: (c) 2010-2013 by the Sentry Team, see AUTHORS for more details.
```
tsoyviktor: All these are `.py` files, I think it would be great to leverage python `date` module and generate valid date.

dcramer: honestly id rather just rm -rf all of these headers

ashwoods:
Just going to kill them all

Issue #1005, opened 2017-05-22
Sanic integration


jrocketfingers:
I assume there are _no_ plans whatsoever to integrate with [Sanic](https://github.com/channelcat/sanic), so I'll be handling it manually, but I'll be more than interested in writing an integration myself.

For additional context, Sanic is an asyncio based framework that's made to go fast™, with an API cloning that of Flask.

I'll begin work on this starting next week, but in the meantime, does any useful advice spring to mind?

dcramer:
@jrocketfingers should be mostly easy as you only need to hook into a common point for catching errors. Outside of that, the Flask integration is probably the best to imitate.

jrocketfingers:
I haven't been able to get to work on this.

For the random internet citizen: in the meantime, you can log to Sentry with something along the lines of:
```
# whever-you-initialized-your-app.py
@app.exception(Exception)
def log_to_sentry(request, exception):
    request.app.raven.captureException({'extra': {'request': request}})
```

antinescience:
@jrocketfingers that PR's ready to go, minus one test I can't reproduce failures on locally.

PR #1205, opened 2018-02-28
Add a client for Sanic.

antinescience:
As mentioned in #1005, it'd be nice to have a client available for folks using the Sanic web server. This introduces that client, along with a set of tests. Considering Sanic's API is very similar to Flask's, the two clients look incredibly similar.

Some questions that arose from doing this:
- In order to ignore a whole lot of logging kruft, I set some default exclusions in the constructor as opposed to [where some of the other defaults are](https://github.com/getsentry/raven-python/blob/master/raven/conf/__init__.py#L14). I have concerns that since Sanic uses `root` as its default logger name, there'd be collisions down the road. Seems safer to scope it to the client as opposed to in defaults.
- As mentioned previously, there are a _lot_ of similarities between the Flask and Sanic clients. I considered refactoring a bit to pull some of the common functionality out, but it seems outside of the scope of this.
- We don't really have an equivalent to Werkzeug/Flask's `request.environ` in Sanic; I took a stab at finding some common variables but I might be missing something simpler.
- Only at 83% coverage. Gonna dig in and figure out where I'm missing things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,9 @@ jobs:
       python: 3.5
       env: TOXENV=py35
     - stage: core
+      python: 3.6
+      env: TOXENV=py36
+    - stage: core
       python: pypy
       env: TOXENV=pypy
     - stage: core
@@ -124,6 +127,14 @@ jobs:
     - stage: contrib
       python: 3.6
       env: TOXENV=py36-lambda
+
+    - stage: contrib
+      python: 3.5
+      env: TOXENV=py35-sanic-07
+
+    - stage: contrib
+      python: 3.6
+      env: TOXENV=py36-sanic-07
 
     - stage: deploy
       script: ./setup.py sdist --formats=gztar bdist_wheel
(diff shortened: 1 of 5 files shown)


antinescience:
@ashwoods looking for a little traction on this one; do you think it'd make more sense for me to bust it out into another repository and support it separately from core raven?

ashwoods:
@antinescience I'd include it for now, just need a day or two more to fix the testing.

antinescience:
Okay! Keep me posted if there's anything I can help out with.

mitsuhiko:
I tried to fix the tests but i have no idea why they fail. @antinescience any clues?

ashwoods:
@mitsuhiko @antinescience provisional rebased fix here: https://github.com/getsentry/raven-python/tree/pr/1205

antinescience:
@mitsuhiko so for me _locally_, they work fine. That's the most perplexing part.

@ashwoods I'll pull the provisional fix down and play with it today.

mitsuhiko:
Yeah. I tried to repro it but I didn't succeed yet.

mitsuhiko:
Merged a fixed version. Thanks so much!

antinescience:
@mitsuhiko nice! Thanks man. I'll definitely open an issue.

Issue #1079, opened 2017-09-14
Add a namedtuple serializer before IterableSerializer


Daenyth:
When I have a `namedtuple` instance value in my stack locals, it gets presented as a list, rather than with its field names. From digging through the code, it looks like adding a namedtuple `Serializer` subclass would allow nicer handling, presenting field names.

Is this reasonable? If I have time to write a PR for this, would it be accepted?

ashwoods:
@Daenyth go for it!

PR #1207, opened 2018-03-08
Namedtuple Serializer

chebee7i:
Addresses #1079.

I didn't quite grok the situations when the `BetterJSONEncoder` would handle the `namedtuples` vs the `NamedtupleSerializer`, but I updated both just in case---it seems somewhat odd that these two paths exists. So if I'm missing something, please let me know.

I did some recent testing on my end by logging an error and passing parameters that contained namedtuples.  I also passed it through the `extra` kwarg. Note that in order to get it working for the `extra` kwarg, I had to apply b588874.  Things seem to be working, but I think it would be best if someone more  familiar with the code could check this branch out and run the full suite of tests.  I've had similar code running in a production environment for ~3 months.  A few minor changes just now, but the core concept is functional.

### DIFF
--- a/raven/utils/__init__.py
+++ b/raven/utils/__init__.py
@@ -15,7 +15,9 @@ except ImportError:
 import sys
 
 from raven.utils.compat import iteritems, string_types
-from raven.utils.basic import merge_dicts, varmap, memoize, once
+from raven.utils.basic import (
+    merge_dicts, varmap, memoize, once, is_namedtuple
+)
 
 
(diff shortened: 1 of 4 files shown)


chebee7i:
Any chance we can get movement on this?

ashwoods:
@chebee7i mind rebasing this against master to get the tests to pass?

chebee7i:
@ashwoods all set!

PR #1213, opened 2018-03-21
Match nodejs client environment variable support

alukach:
I was surprised to find that the Python client does not support the same environment variables [documented in the NodeJS client](https://docs.sentry.io/clients/node/config/#environment-variables).  This is a quick fix so maybe there's a particular reason that this choice was made that I'm not aware of.  However, I'd recommend adding support for these variables.

This PR is mostly about kicking the tires on the idea.  If this is something that we're interested in merging, I can throw together some quick tests to verify that these remain effective for time to come.


alukach:
i'm a little bit lost as to why I'm getting slapped so hard on these tests, any tips?

ashwoods:
@alukach can you rebase from master?

alukach:
@ashwoods done, thanks for the pointer.

ashwoods:
thx!

PR #1218, opened 2018-03-29
feat: Disable dill logger by default

dcramer:


dcramer:
Someone can educate me on what dill is from, but we see it in Sentry so I think its a safe default (as its not us).

mattrobenolt (approved):

PR #1233, opened 2018-05-02
Add blinker to Sanic dependencies.

antinescience:
Pulling 6.7.0 and trying to do, `from raven.contrib.sanic import Sentry`, I get the following traceback:

```python
  File "application.py", line 1, in <module>
    from raven.contrib.sanic import Sentry
  File "/usr/local/lib/python3.6/site-packages/raven/contrib/sanic.py", line 13, in <module>
    import blinker
ModuleNotFoundError: No module named 'blinker'
```

That's my bad. This should resolve that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,7 @@ if sys.version_info[0] == 3:
 if sys.version_info >= (3, 5):
     sanic_requires = [
         'blinker>=1.1',
-        'sanic>=0.7.0', 
+        'sanic>=0.7.0',
     ]
     sanic_tests_requires = ['aiohttp', ]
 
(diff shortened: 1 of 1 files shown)


ashwoods:
@antinescience no worries. small flake8 error: https://travis-ci.org/getsentry/raven-python/jobs/374025469#L745

TheKevJames:
I think I'm misunderstanding something about this PR; as far as we can tell, `sanic_requires` is not installed when running `pip install raven`. Should we add `'sanic': sanic_requires,` to the `extras_require` key, as we do with Flask?

Issue #1258, opened 2018-06-07
`from raven import Client` takes 17MB of RAM


skrap:
`from raven import Client` takes 17MB of RAM

Assumptions:
* swapless linux system (embedded in my case, but would also affect cloud)
* python 2.7

To reproduce:
```
:/tmp# python
Python 2.7.14 (default, Jun  6 2018, 06:11:27) 
[GCC 6.4.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import psutil
>>> psutil.Process().memory_info().rss
7974912
>>> from raven import Client
>>> psutil.Process().memory_info().rss
25337856
>>> 25337856-7974912
17362944
```

The numbers may vary slightly from system to system, but you get the idea.

This comment, from `raven.transport.registry`, seems relevant:
```
# TODO(dcramer): we really should need to import all of these by default
```
(I think dcramer probably meant "should not")

untitaker:
Hi @skrap, this should no longer be a problem with the new SDK which only has a single transport implemented: https://github.com/getsentry/sentry-python

PR #1260, opened 2018-06-11
Default usage of raven.Client does not import twisted or requests

skrap:
This patch addresses #1258 by taking steps to lazily load the requests and twisted modules.

Summary of changes:
* Delay the `hook_libraries` function if the module requested hasn't been loaded yet.
* Remove and instead lazy load DEFAULT_TRANSPORT (wasn't used outside its .py).
* Remove the default loading of `twisted` and `requests` in their respective transport subclasses.  Lazy load what's needed.

Caveats:
* While I modified existing tests to pass, I was not sure how to write a unit test for this, as performance tests can be tricky.
* While this should work in all supported pythons, I'm only able to run the tests in latest 2- and 3- series.

### DIFF
--- a/raven/breadcrumbs.py
+++ b/raven/breadcrumbs.py
@@ -1,7 +1,13 @@
 from __future__ import absolute_import
 
+try:
+    import __builtin__ as builtins
+except ImportError:
+    import builtins
+
 import os
 import logging
+import sys
 
 from time import time
 from types import FunctionType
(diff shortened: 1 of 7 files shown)


skrap:
Build failures seem unrelated to these changes:
```pkg_resources.VersionConflict: (pytest 3.2.5 (/home/travis/build/getsentry/raven-python/.tox/py35-django-200-fix/lib/python3.5/site-packages), Requirement.parse('pytest>=3.6'))```

ashwoods:
Yup, almost have a fix ready

ashwoods:
@skrap fixed, then broke again,then fixed again. Mind rebasing from master?

PR #1261, opened 2018-06-14
fix: Fix stacktraces in some situations being the wong way round

mitsuhiko:
Oddly enough even the tests were wrong.


mitsuhiko:
@mattrobenolt as far as I can tell this was wrong forever but since the tests were also incorrect and it's late here I want a second pair of eyes that have seen this code before.

mattrobenolt:
Yeah, this code here is only used when generating a synthetic stacktrace, right? When we have a real traceback, we do the right thing.

Issue #922, opened 2016-11-28
Getting Celery logging via Django into Sentry


Duologic:
Repost from https://github.com/getsentry/sentry/issues/4565 on request of @JTCunning.

Recently we updated Celery to 4.0 in our Django project. The logging gave us some headaches and I would like to help fellow hackers so they don't get stuck on this.

The issue is described in the Celery issue tracker but disregarded as they don't have enough resources to work on it: https://github.com/celery/celery/issues/2437

We had to implement 2 things:

Set `CELERYD_HIJACK_ROOT_LOGGER` to `False` in our settings.

And connect to the `setup_logging` signal from Celery:

```
from celery import signals

@signals.setup_logging.connect
def setup_logging(**kwargs):
    """Setup logging."""     
    pass
```

It would be nice to see this reflected in the documentation somewhere. As it looks to me this is very Django-specific, I'd propose to putting it here: https://docs.sentry.io/clients/python/integrations/django/

I also got confused from the settings described on the page about Celery in the Sentry docs. The code there doesn't seem to be required for us. Not sure if we can change something about this to make it more clear when you should use that. Only when you are running celery<4.0 maybe? https://docs.sentry.io/clients/python/integrations/celery/

Otherwise, the docs are great, thank you for that!

What do you guys think?

pabdelhay:
Hey @Duologic , 
Where did you put this code from setup_logging? In **celery.py** (do you config celery here?)?
Thanks in advance

Duologic:
Wow, that is from quite some time ago. I'm not sure if it still works that way, but looking at my own description, that looks like the place to put it. As the decorator makes sure the signal is connected, I'd think it doesn't matter, as long as it gets executed when celery is set up.

Duologic:
As this SDK is being phased out for Sentry-Python, no need to keep this issue open.

PR #1267, opened 2018-06-29
Fix celery error logging

tilboerner:
Error logging from Celery tasks is broken out of the box. While uncaught exception reporting works, error-level log messages from inside Celery worker do not get reported to Sentry. 

For the user, it's not obvious that this is happening, or what can be done to fix it.The setup instructions for [Django][] and [Celery][] are not sufficient to get it working. There is documentation on [raven.readthedocs.io][] suggesting that the Django handler "captures the errors" from workers. Consequently, there is #922, suggesting a workaround.

[Django]: https://docs.sentry.io/clients/python/integrations/django/
[Celery]: https://docs.sentry.io/clients/python/integrations/celery
[raven.readthedocs.io]: https://raven.readthedocs.io/en/stable/integrations/django.html#additional-settings

The problem is that when setting up logging for *worker processes*, Celery configures the task logger to not propagate to the root logger:

https://github.com/celery/celery/blob/47ca2b462f22a8d48ed8d80c2f9bf8b9dc4a4de6/celery/app/log.py#L132

https://github.com/celery/celery/blob/47ca2b462f22a8d48ed8d80c2f9bf8b9dc4a4de6/celery/app/log.py#L176

This prevents log events from reaching the Celery handler in the root logger. (Note that this setup does not happen in the Django shell, for example. There, things work fine.)

The workaround in #922 disables this logging setup completely. I'd argue that it's cleaner to work with whatever Celery decides to do and not mess with the logging system any further, beyond adding sentry handlers.

So, this PR adds a signal handler for the task logger, to give it the sentry handler if needed. It also adds code so that the `DjangoCeleryHandler` sets this up properly out of the box, if any of the `CELERY_LOGLEVEL` settings are defined. (This way it's opt-in; it could also be enabled by default.)

People should not have to discover the hard way that their errors do not properly get logged, so I think Raven should really fix this and work out of the box. If any manual action remains to be required, this should be noted in the documentation.

If you're willing to consider this PR, I'd be happy to add tests and documentation. In that case, it would be great if you could point me towards the proper places to put them.


tilboerner:
@ashwoods Rebased it. Could you restart that one failed job? Looks like some pip download timed out.

blodone:
its really really painful to enable logging celery to sentry... +1

huangsam:
Is there anything blocking this from getting released? This PR would be incredibly useful for folks who need to track Celery errors via the Raven client.

untitaker:
To all, I think the proper fix is to upgrade to the new [Python SDK](https://github.com/getsentry/sentry-python) which bypasses those problems entirely.

I'm a bit vary of fixing this in raven as it is a breaking change for something that is effectively in maintenance mode.

Issue #1155, opened 2017-12-06
'encoding' cannot be specified in python 3


jonapich:
https://github.com/getsentry/raven-python/blob/1a0d697dd11459bfa8ce933b3b53264098dad60d/raven/utils/json.py#L50

The 'json.dumps' command no longer have an encoding argument in python 3 and causes this stacktrace:

```
File "/usr/local/lib/python3.6/site-packages/raven/handlers/logging.py", line 89, in emit
 return self.emit(record)
 File "/usr/local/lib/python3.6/site-packages/raven/handlers/logging.py", line 182, in _emit
 *kwargs)
 File "/usr/local/lib/python3.6/site-packages/raven/base.py", line 645, in capture
 self.send(*data)
 File "/usr/local/lib/python3.6/site-packages/raven/base.py", line 740, in send
 message = self.encode(data)
 File "/usr/local/lib/python3.6/site-packages/raven/base.py", line 782, in encode
 return zlib.compress(json.dumps(data).encode('utf8'))
 File "/usr/local/lib/python3.6/site-packages/raven/utils/json.py", line 51, in dumps
 return json.dumps(value, cls=BetterJSONEncoder, **kwargs)
 File "/usr/local/lib/python3.6/json/init.py", line 238, in dumps
 **kw).encode(obj)
TypeError: __init_() got an unexpected keyword argument 'encoding'
```

ashwoods:
@jonapich I'm having problems trying to create a test case for this issue. Have an example of sentry data where this exception is raised?

jonapich:
No I don't. I'm not sure a sentry was even raised for this, it probably ended up in our logs instead. I guess there was some invalid UTF data in the exception/locals somewhere?

cenkalti:
@ashwoods we experience this too. The code puts `encoding` kwarg into `json.dumps` call but it is not valid for Python 3. If this code path is executed, it will always fail on Python 3. We see this frequently on Sentry. Here is an example traceback: https://cl.ly/okPZ

Let me know if you need anything else for debugging.

miracle2k:
I don't think this should be closed.

I am seeing this still, and it's a bit confusing, but here is what is happening. We have this code in Sentry:

https://github.com/getsentry/raven-python/blob/1a0d697dd11459bfa8ce933b3b53264098dad60d/raven/utils/json.py#L46-L51

[...]

PR #1273, opened 2018-07-14
Ensure that all keys in locals are strings before serializing

brakhane:
Trio uses an ugly hack and puts a value into locals with a
non string key (class LOCALS_KI_PROTECTION_ENABLED).

When trying to serialize that, JsonEncoder crashes.
So we ensure that every key is indeed a string.

Another way to solve this problem would be to pass skipkeys=True
to json.dumps, but this might mask other errors.

### DIFF
--- a/raven/utils/stacks.py
+++ b/raven/utils/stacks.py
@@ -14,8 +14,7 @@ import os
 import sys
 
 from raven.utils.serializer import transform
-from raven.utils.compat import iteritems
-
+from raven.utils.compat import iteritems, string_types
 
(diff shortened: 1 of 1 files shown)


brakhane:
This took really long to find, because of #1155 masking the real error, BTW. It would be nice if an exception during serialization wouldn't fail with a confusing message on Python3.

Here's the line in trio that adds a guard object into locals:
https://github.com/python-trio/trio/blob/79d66647c3c947f614b51f607be6fd527df6777f/trio/_core/_run.py#L1245

PR #1279, opened 2018-07-30
Update processors.py

### LINKED ISSUE #1253: SanitizePasswordsProcessor and local vars
I'm experiencing an issue where values in certain locations are being masked by the `SanitizePasswordsProcessor` and not in others, namely in local variables pulled from stack traces.

The default regex for the processor is `^(?:\d[ -]*?){13,16}$` and when I have code such as this in my app:
```
@app.route('/')
def index():
    # masked correctly
    current_app.sentry.client.context.merge({'extra': {
        'example': '4242424242424242'}})

    # shows up in sentry wrapped in single quotes
    example_var = "4242424242424242"
    raise Exception('foo')
```

the example from the extra context is masked correctly with asterisks, and the example_var from the stack trace is in plain text with quotes around it. It appears when it enters the sanitize method it is a string wrapped in single quotes which fails `SanitizePasswordsProcessor.VALUES_RE`.

Version information:
```
Python 3.6.5
raven==6.8.0
```

Is this expected behavior?

ehfeng:
Current regex does not match against credit card numbers in the stack local values, because they are wrapped in single quotes: `'4242424242424242'`. It could be that credit card numbers should always be represented as integers, but it seems harmless to catch this case as well.

Fixes #1253

### DIFF
--- a/raven/processors.py
+++ b/raven/processors.py
@@ -162,7 +162,7 @@ class SanitizePasswordsProcessor(SanitizeKeysProcessor):
         'sentry_dsn',
         'access_token',
     ])
-    VALUES_RE = re.compile(r'^(?:\d[ -]*?){13,16}$')
+    VALUES_RE = re.compile(r'^\'?(?:\d[ -]*?){13,16}\'?$')
 
     @property
     def sanitize_keys(self):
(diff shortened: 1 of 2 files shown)


153957:
Can you add a test which verifies that this fixes #1253?

untitaker:
I'm closing this because we need something way more generic that checks on arbitrary word boundaries. Also we don't want to have this kind of logic in SDKs in the future anyways.

PR #1280, opened 2018-07-31
Support longer credit card numbers, upto and including 19 digits.

153957:
According to multiple sources credit cards can now be 19 digits long.

- https://www.mastercard.us/content/dam/mccom/global/documents/mastercard-rules.pdf
    > For a Maestro Account, the PAN must be 13 to 19 digits in length.
- https://www.freeformatter.com/credit-card-number-generator-validator.html
- https://en.wikipedia.org/wiki/Payment_card_number


153957:
This repo is legacy, so nevermind.

PR #1299, opened 2018-09-11
setup.py: add extra target for Sanic dependencies

acdha:
This ensures that someone installing sanic support will get the blinker dependency as well

See #1005


iScrE4m:
Merge would be greatly appreciated!

untitaker:
Thanks!

PR #1300, opened 2018-09-14
Use correct kwarg in handle_exception() for Flask

ltm:
The handle_exception() method is registered as a receiver of the
got_request_exception signal. According to both the [documentation][1]
and [source code][2] of Flask the got_request_exception signal passes
the exception as `exception` rather than a `exc_info`.

This is likely to have gone unnoticed since captureException() calls
sys.exc_info() in the absence of an exception.

On Python 3 we can use `__traceback__` to explicitly construct the
`exc_info`.

[1]: http://flask.pocoo.org/docs/1.0/api/#flask.got_request_exception
[2]: https://github.com/pallets/flask/blob/1.0.2/flask/app.py#L1732

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ------
 
 * [Core] Fixed stackframes in some situations being in inverse order.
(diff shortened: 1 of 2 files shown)


untitaker:
thanks!

PR #1301, opened 2018-09-14
remove docs link to nowhere

Carthaca:
With ea000f57b6f249974293003e97b84741323fadf5 docs folder was removed, therefore the pre-commit hook should go to.


markstory:
Thank you 👍

PR #1306, opened 2018-10-08
Fix sanitizing of passwords with "=" symbol

### LINKED ISSUE #1263: password not filtered when it contains `=`
We noticed getting unfiltered passwords into our Sentry.

The environment we run is:

- Python==3.6.4
- Django==2.0.2
- djangorestframework==3.7.7
- raven==6.9.0

With following json in the body of the call
```
{
	"password":"blabla",
	"code":"12345"
}
```
we got in Sentry
Body
```
{
code: test1, 
password: [Filtered]
}
```
So far so good

But now come the strange thins

### Strange Example 1
```
{
	"password":"blablabla",
	"code":"=lookthistext&testb"
}
```
gives:

Body
```
{
code: =********&testb, 
password: [Filtered]
}
```
Password is filter, but already strange behaviour

### Strange Example 2
```
{
	"password":"bla=blabla",
	"code":"test3"
}
```
gives:

Body
```
{ "password":"bla | [Filtered]
```
So the password before `=` is visible

### Strange Example 3
```
{
	"password":"blablablabla-test3",
	"code":"test=123-456"
}
```
gives:

Body
```
{ "password":"blablablabla-test3", "code":"test | [Filtered]
```
So the complete password is visible

Is this missing settings or a bug in Raven?

orsinium:
1. Splited HTTP filtering by separated processors for HTTP query and mapping
1. Added JSON processing to HTTP filtering. Close #1263
1. Included test case from original issue.

### DIFF
--- a/raven/processors.py
+++ b/raven/processors.py
@@ -7,6 +7,7 @@ raven.core.processors
 """
 from __future__ import absolute_import
 
+import json
 import re
 import warnings
 
@@ -107,6 +108,11 @@ class SanitizeKeysProcessor(Processor):
             frame['vars'] = varmap(self.sanitize, frame['vars'])
 
     def filter_http(self, data):
+        processors = (
+            self._process_json,
+            self._process_http_query,
+            self._process_dict,
+        )
         for n in ('data', 'cookies', 'headers', 'env', 'query_string'):
             if n not in data:
                 continue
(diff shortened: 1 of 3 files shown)


orsinium:
@dcramer, could you, please, review this PR?

orsinium:
Is anybody here?

untitaker:
Since you started your pr we have written a new python sdk from scratch which purposefully does not try to replicate ravens logic but rather only gives you the hooks to inject your own code that may apply regexes.

https://github.com/getsentry/sentry-python

orsinium:
Thank you for your reply

PR #1309, opened 2018-11-02
Allow `extra` to be None in breadcrumbs processing

joker-official:
Sometimes, `extra` is passed in as a default parameter of another
function and can therefore be `None`. This fixes:

	  File "raven/breadcrumbs.py", line 77, in get_buffer
		processor(payload)
	  File "raven/breadcrumbs.py", line 143, in processor
		data_value.update(extra)
	TypeError: 'NoneType' object is not iterable

### DIFF
--- a/raven/breadcrumbs.py
+++ b/raven/breadcrumbs.py
@@ -141,6 +141,8 @@ def _record_log_breadcrumb(logger, level, msg, *args, **kwargs):
 
         # Extract 'extra' key from kwargs and merge into data
         extra = kwargs.pop('extra', {})
+        if extra is None:
+            extra = {}
         data_value = kwargs
         data_value.update(extra)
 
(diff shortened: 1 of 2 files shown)


david-volantio:
The issue addressed by this PR is preventing my team from upgrading to 6.10.0. The issue in my particular project arises when calling `captureMessage` (that I know of - I haven't exhaustively testing every call in my project) without `extra`, so I'm not sure where `extra` is being defaulted to `None` but it would seem to be between `captureMessage` and breadcrumb processing.

ashwoods:
@david-volantio if you are able to upgrading, any reason for not being able to give the new sdk a go? https://github.com/getsentry/sentry-python

david-volantio:
@ashwoods a fair point - I was just hoping we could move to the latest release but updating to sentry-python is definitely on our roadmap. thanks!

PR #1312, opened 2018-12-14
fix: Dont serialize nan when present in frame local vars

untitaker:
We want to use Rust on the server to decode JSON and it's not possible to just hook into there and add support for new constants (essentially new syntax) without forking the JSON library or running some code before deserializing.

The exact values are how `frame.vars` are coerced on the server anyway so there is no difference in display to the user.

Raven still permits sending NaN as part of other variables. Sentry-sdk (in master) treats it like any unserializable value. Serverside this is converted to `None` while deserializing.

Not sure if coercing to string would be that great (might create new groups) or if we should rather just crash in the SDK like sentry-sdk does now and drop the event.

### DIFF
--- a/raven/utils/serializer/base.py
+++ b/raven/utils/serializer/base.py
@@ -174,17 +174,8 @@ class FloatSerializer(Serializer):
     types = (float,)
 
     def serialize(self, value, **kwargs):
-        value = float(value)
-
-        if value == float('inf'):
-            return '<inf>'
-        if value == float('-inf'):
-            return '<-inf>'
-        # lol checking for float('nan')
-        if value != value:
-            return '<nan>'
-
-        return value
+        # Wrap with repr to convert inf/nan to string
+        return repr(float(value))
 
 
 class IntegerSerializer(Serializer):
--- a/raven/utils/serializer/base.py
+++ b/raven/utils/serializer/base.py
@@ -174,7 +174,17 @@ class FloatSerializer(Serializer):
     types = (float,)
 
     def serialize(self, value, **kwargs):
-        return float(value)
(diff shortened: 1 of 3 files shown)


raven/utils/serializer/base.py
```diff
@@ -174,7 +174,17 @@ class FloatSerializer(Serializer):
     types = (float,)
 
     def serialize(self, value, **kwargs):
-        return float(value)
+        value = float(value)
+
+        if value == float('inf'):
```
dcramer: Given this is a repr serializer, why don't we return what Python would actually render.

For example:


```python
>>> repr(float('-inf'))
'-inf'
```

untitaker: I tried to preserve the behavior that Sentry uses to normalize infinity/nan, but thinking about it there's no way this would actually break stuff. Will fix.

raven/utils/serializer/base.py
```diff
@@ -174,7 +174,8 @@ class FloatSerializer(Serializer):
     types = (float,)
 
     def serialize(self, value, **kwargs):
-        return float(value)
+        # Wrap with repr to convert inf/nan to string
+        return repr(float(value))
```
dcramer: Should we also normalize Ints/Longs/etc?

untitaker: I was trying to keep this patch as minimal as possible but we can do that too

untitaker:
So, every number is a string now. I think this is not necessary for dicts and lists as the trimming behavior is more sophisticated there. We might actually want to port parts of this from raven to the new sdk

PR #1350, opened 2019-03-18
Fix #1349 Skipkeys when serializing to json

### LINKED ISSUE #1349: Use `skipkeys` when serializing payload
See comment https://github.com/getsentry/raven-python/issues/1155#issuecomment-473423879

When serializing error data (traceback, local vars, etc.) to be sent to the backend, `json.dumps` is used, in `raven.utils.json`. `json.dumps` can fail, e.g. when local data has the form:
```
{ (): "foo" }
```
which cannot be serialized to json (because of the tuple key).
By using `json.dumps(skipkeys=True)` these kind of keys will be ignored.

lbolla:
Also, don't try to call json.dumps with "encoding" keyword in Python3, which
does not support it.

### DIFF
--- a/raven/utils/json.py
+++ b/raven/utils/json.py
@@ -15,6 +15,7 @@ import uuid
 import json
 
 from .basic import is_namedtuple
+from .compat import PY2
 
 
 try:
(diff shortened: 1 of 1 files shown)


lbolla:
The above TravisCI failures don't seem to have anything to do with this patch.

PR #1363, opened 2020-04-01
Zh 5.14.0 override hooks

zhammer:


untitaker:
I assume you didn't want to submit this patch against raven-python. You can use the raven client options to turn off this behavior instead of disabling the code like that.

